PR #6, opened 2012-12-10
counter_culture_fix_counts not respecting deleted_at columns

kiere:
When inspecting some of the counter cache column counts on our dev machines, and then the query generated by counter_culture_fix_counts, we noticed it wasn't picking up our deleted_at columns for paranoid deletes.

### DIFF
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -74,6 +74,8 @@ module CounterCulture
           # we are only interested in the id and the count of related objects (that's this class itself)
           query = klass.select("#{klass.table_name}.id, COUNT(#{self.table_name}.id) AS count")
           query = query.group("#{klass.table_name}.id")
+          # respect the deleted_at column if it exists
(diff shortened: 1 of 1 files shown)


voidless:
This should be optional.
I need to disallow editing some fields of objects that have related objects even if they are soft deleted.
This cannot be achieved with this change so we have to use rails counter_cache for objects like this and it produces incorrect counts.

This could be done like this instead of hardcoded check for deleted_at in the gem:
`counter_culture :category, :column_name => proc {|model| model.deleted_at.blank? ? "products_count" : nil }`
or it can be optional with a parameter like that:
`counter_culture :category, count_deleted: false`

magnusvk:
Feel free to open a pull request to add the `count_deleted` flag. I'd be open to merging that. I don't like the idea of ripping this out after it's been the standard behavior for ~3.5 years.

I don't understand enough about your usage here to be sure, but it also sounds like you might be able to work around this by adding a separate custom state to these records that is not "deleted" but still hides them from the default scope.

Issue #91, opened 2015-01-20
delta_column not accepting updates when enum column


sebastianhoitz:
I have an enum column called `score` which maps score values to human-readable attributes.

This works well when using the counter_culture_fix_counts method because it uses the values provided in the database.

However, when using the after_update hook, counter_culture wants to update the value with a query like this:

```
UPDATE "procedures" SET "total_score" = COALESCE("total_score", 0) + essential
```

The expected behavior would be to use the enum value here: https://github.com/magnusvk/counter_culture/blob/master/lib/counter_culture.rb#L287

```ruby
        delta_magnitude = if options[:delta_column]
                            delta_attr_name = options[:was] ? "#{options[:delta_column]}_was" : options[:delta_column]
                            value = self.send(delta_attr_name) || 0
                            value.is_a? Enumerize::Value ? value.value : value
                          else
                            1
                          end
```

sebastianhoitz:
Closed since I opened a pull request.

PR #92, opened 2015-01-20
Add support for rails enum

sebastianhoitz:
This pull requests fixes issue #91.

It basically adds support for enum columns that define human readable names for internally used integer-values. I added a unit test to verify the correct behavior.

### DIFF
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -286,7 +286,12 @@ module CounterCulture
       if id_to_change && options[:counter_column]
         delta_magnitude = if options[:delta_column]
                             delta_attr_name = options[:was] ? "#{options[:delta_column]}_was" : options[:delta_column]
-                            self.send(delta_attr_name) || 0
+                            val = self.send(delta_attr_name)
+                            # check if value is a string. This can happen if the
+                            # delta_column is an enum
+                            val = val.is_a?(String) ? self[delta_attr_name.to_sym] : val
+
+                            val || 0
                           else
                             1
                           end
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -289,8 +289,7 @@ module CounterCulture
                             val = self.send(delta_attr_name)
                             # check if value is a string. This can happen if the
                             # delta_column is an enum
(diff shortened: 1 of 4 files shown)


lib/counter_culture.rb
```diff
@@ -286,7 +286,11 @@ def change_counter_cache(options)
       if id_to_change && options[:counter_column]
         delta_magnitude = if options[:delta_column]
                             delta_attr_name = options[:was] ? "#{options[:delta_column]}_was" : options[:delta_column]
-                            self.send(delta_attr_name) || 0
+                            val = self.send(delta_attr_name)
+                            # check if value is a string. This can happen if the
+                            # delta_column is an enum
+                            val = val.is_a?(String) ? self.class.send(options[:delta_column].pluralize)[val] : val
```
magnusvk: I'm not a big fan of checking for `String` here. Is there a way to more explicitly check whether the column is an enum and then explicitly unbox the value? This seems a little weird and brittle to me, in all honesty. You had something in the issue you had opened before that seems closer to something I'd want to do here.

magnusvk:
Happy to include something like this in principle though.

lib/counter_culture.rb
sebastianhoitz: In the issue I opened I still had the `enumerize` gem installed, that's why there was this special `Enumerize::Value` object. I realized this when writing the unit test and the test failed to find the object Enumerize.

I feel the same as you.

Sadly, Rails does not really provide ways to actually detect if the column is an enum. It just adds a bunch of methods that allow you to work with it, see here: http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

So what we could do to make a better guess is:

* Check if self.class.send(options[:delta_column].pluralize) returns a Hash

Any further ideas?

magnusvk: What does `read_attribute` return for an enum? With a little luck that returns the integer value. :) Then we could just do `self.read_attribute(delta_attr_name)` instead of the `send`.

magnusvk:
Closing this for lack of response. Feel free to reopen.

Issue #93, opened 2015-02-10
does not work correctly on relations with foreign_key option


jogaco:
If I have:

```
class AnotherPost < ActiveRecord::Base
  has_many :comments, class_name: 'AnotherPostComment', foreign_key: 'another_post_id', primary_key: 'another_id'
end

class AnotherPostComment < ActiveRecord::Base
  belongs_to :post, class_name: 'AnotherPost', foreign_key: 'another_post_id', primary_key: 'another_id'
  counter_culture :post
end
```
it wrongly updates the counter cache searching by id.

PR #94, opened 2015-02-10
Fix for #93: does not work correctly on relations with foreign_key option

jogaco:


magnusvk:
Merged and released as 0.1.30. Thank you!

jogaco:
Great!

It is also missing the handling of conditions option, but it gets too
difficult to follow with the relationship loop in the source and I needed
my fix asap for a bug in production.
El dia 10/02/2015 16.46, "Magnus von Koeller" <notifications@github.com> va
escriure:

PR #104, opened 2015-04-24
Fix behavior with update and destroy trigger in the same transaction

biomancer:
I've run into case when there were multiple updates and then destroy in the same transaction, and the counter was not updated correctly as only update callback was scheduled and it did not affect counter, while the destroy callback was declined. I've updated code to differentiate between create/update and destroy callbacks when scheduling them.
I cannot imagine any case, when this would be undesired behavior, so there it is.


magnusvk:
Ok; so I don't quite know what this is fixing. Is there a way you can add a regression spec, too? That would help me understanding why this is necessary, too. Thanks for the contribution.

PR #107, opened 2015-06-19
update readme about gem installation instructions

jaimeiniesta:
Running `bundle update` would update all the gems in the application, better run just `bundle install` to install this particular gem.

I've also updated the version number, but I'd just leave it out entirely and use `gem 'counter_culture'` to install the latest version.


magnusvk:
Jaime -- thank you. I'd like to keep the version number, as that makes sure that any future API incompatible versions won't get pulled in automatically. As long as the releases are just at the patch level, it will still pull in the latest version even if you'd use an outdated version with the tilde arrow.

PR #114, opened 2015-10-27
`Ambiguous column name id` fix

ProGM:
If the counting table has a default_scope that contains a JOIN, counter_culture crashes with an error like this:
```
Ambiguous column name: id: SELECT  DISTINCT "users".* FROM "users" LEFT OUTER JOIN companies WHERE (id = 53)  ORDER BY "users"."id" ASC LIMIT 1
```

I added some tests and a fix. ;)

### DIFF
--- a/spec/models/user.rb
+++ b/spec/models/user.rb
@@ -9,21 +9,21 @@ class User < ActiveRecord::Base
   has_many :reviews
   accepts_nested_attributes_for :reviews, :allow_destroy => true
 
-  default_scope {
-    if self._default_scope_enabled
+  default_scope do
+    if _default_scope_enabled
       joins("LEFT OUTER JOIN companies").uniq
     else
       all
     end
-  }
-
-  def self._default_scope_enabled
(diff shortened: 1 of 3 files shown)


spec/models/user.rb
```diff
@@ -8,4 +8,22 @@ class User < ActiveRecord::Base
 
   has_many :reviews
   accepts_nested_attributes_for :reviews, :allow_destroy => true
+
+  default_scope {
```
deniskorobicyn: Maybe it'll be better rewrite it more in ruby way?
```ruby
default_scope do
  ...
end
```

ProGM: Fixed :+1: 

... and improved a few minor things

deniskorobicyn: nice! :+1:

magnusvk:
:+1: thank you for this.

magnusvk:
Released as 0.1.34. Thanks again for contributing.

PR #126, opened 2016-04-20
Refactor to reduce method pollution in AR and better clarity

adzap:
I'm not sure if you're interested, but I have begun a refactor to better encapsulate behaviour and reduce the methods added ActiveRecord.

Also I hope it gives better clarity for what is going on. I had a hard time following it all in large methods. It is still not as clear and intention revealing as I hope to get it.

I will understand if this is too much of a patch bomb but let me know your thoughts.


lib/counter_culture/helpers.rb
```diff
@@ -0,0 +1,138 @@
+module CounterCulture
+  module Helpers
```
magnusvk: I'm not a fan of this name. I would probably call this `ActiveRecordExtensions` to make clear that this gets included into `ActiveRecord::Base`. `Helpers` makes it sound like it's just internally used when the opposite is the case.

Gemfile
```diff
@@ -14,6 +14,7 @@ group :development, :test do
   gem "rails", '>= 3.1.0'
   gem "rspec", "~> 2.10.0"
   gem "awesome_print"
+  gem "byebug"
```
magnusvk: Adding byebug seems to have broken the Ruby 1.9.3 build. Can you look into fixing that?

spec/counter_culture_spec.rb
```diff
@@ -1371,7 +1371,8 @@
     categ.reload.posts_count.should == 1
   end
 
-  describe "#previous_model" do
+  # TODO check that effect is tested rather than method directly
```
magnusvk: Why?

lib/counter_culture/helpers.rb
```diff
@@ -0,0 +1,138 @@
...
+      #   { :exclude => list of relations to skip when fixing counts,
+      #     :only => only these relations will have their counts fixed }
+      # returns: a list of fixed record as an array of hashes of the form:
+      #   { :entity => which model the count was fixed on,
+      #     :id => the id of the model that had the incorrect count,
+      #     :what => which column contained the incorrect count,
+      #     :wrong => the previously saved, incorrect count,
+      #     :right => the newly fixed, correct count }
+      #
+      def counter_culture_fix_counts(options = {})
```
magnusvk: I'm curious: This now partially delegates to the `Reconciler`. Why not move _all_ of this logic over there and just delegate from here?

We could even go a step further and deprecate this method in favor of calling `CounterCulture::Reconciler` directly.

magnusvk:
I like the thrust here. I'd want to spend a little more time thinking about this before hitting merge, but thank you for doing this. I agree this code could use some refactoring.

Question: Were you doing this just to improve the code? Or were you planning to also add functionality / change behavior?

lib/counter_culture/helpers.rb
adzap: Yeah fair point. Given it is an exclusive AR gem, how about just Extensions?

Gemfile
adzap: Will remove for now.

spec/counter_culture_spec.rb
adzap: I haven't gone back to check that the effect of needing the previous model in the dynamic case is handled in the specs. If so I will just remove that spec.

lib/counter_culture/helpers.rb
adzap: I wanted to have more precise control rather than a bulk loop. But you're right It doesn't need to live here. Though Reconciler is the wrong place I think. 

I was originally going to make the array of counters a first class object. But there wasn't enough behaviour to warrant it as it just loops mainly. But moving this method that acts on the collection into a collection object of counters would be my preference.

adzap:
I needed this behaviour in my projects but was put off by the number of methods added to AR and the large impenetrable methods. In my projects I usually find I need to use discrete behaviour at some point for more flexible use, and big looping methods prevent that. So I also wanted to decompose and encapsulate.

I also found the dynamic column config jarring and it wasn't what I was expecting to find when using the gem. So I was also planning to work on that to remove some of the double handling (:column_name => proc and :column_names if you need to run the fix counts). I know that may not work in all cases but I wanted to explore it.

Basically it was either write my own version or adapt this gem. But given I was going to need to reimplement most of it and probably steal a lot of your code if I reimplemented, it made sense to work on the gem.

Ultimately I read the code of most/all gems before I add them to my projects to make sure it ticks a few necessary boxes and avoid regrets.

adzap:
I pushed a few more changes.

My majority use case is a conditional counter change on a static column name. At the moment the dynamic column code allows for any random column name change. I am hoping have something like

`counter_culture :parent, column_name: 'active_child_count', conditions: { active: true }` 

where the conditions could be used for single change or bulk fix.

After all not being able to do both is really untenable. You need to be able to ensure that you could run a bulk fix when adding to existing projects and just in case you need to reset the counts. So that means that you always need an arel compatible where clause(s) as per the :condition_names for any 'dynamic' column. So I am wondering can we just use the where clause and drop the proc altogether.

adzap:
I forgot to say that the conditions would need to be executed as method calls on the single change case.

lib/counter_culture.rb
```diff
@@ -6,6 +6,8 @@
 require 'counter_culture/reconciler'
 
 module CounterCulture
+  mattr_accessor :batch_size
+  self.batch_size = 1000
```
magnusvk: This is great. Can you add some documentation for this to the README?

magnusvk:
I'm not going to drop support for `Proc`s for column names. It gives you a lot of flexibility and works very well for me in my own projects. But I do get that it's not the friendliest syntax, whereas something like `conditions: { active: true}` is a lot simpler to learn, so I'd be happy to add that syntax in addition.

If I were to implement this, I would probably just convert the `conditions` to a `Proc` and `column_names` combination behind the scenes so you don't have to re-implement that logic.

lib/counter_culture.rb
```diff
@@ -1,410 +1,14 @@
...
-      self.class.send :first_level_relation_foreign_key, relation
-    end
-
-  end
-
-  # extend ActiveRecord with our own code here
-  ::ActiveRecord::Base.send :include, ActiveRecord
+module CounterCulture
+  mattr_accessor :batch_size
+  self.batch_size = 1000
```
magnusvk: This is great. Can you add some documentation for this to the README?

spec/counter_culture_spec.rb
magnusvk: I would want to keep this spec regardless. This method is supposed to implement specific behavior and I'd want to keep a test to ensure it actually works correctly, even if it's indirectly tested elsewhere.

magnusvk:
Thanks for doing this! I think this is looking great. My inclination would be to merge this before you add any new / changed behavior, as a new baseline in master. Let me know when you think it's ready.

adzap:
Sounds great. I am done now. Added some documentation.

I changed the config to be CounterCulture.config.batch_size where config just returns the CounterCulture module for now. This is to allow a simple way to replace it later with a proper configuration object but doesn't need any app code changes.

magnusvk:
👍 thanks again for doing this!

magnusvk:
Released as 0.2.0.

adzap:
No worries. Thanks for the gem.

PR #127, opened 2016-04-24
Implement specific increment/decrement counter methods

adzap:
I found the change_counter_cache is a little too open for a public method so I implemented AR-style increment/decrement methods which is more intentional and less error prone.

Also a typo fix, spec fix, and moved an option validation into the counter class in this PR as well, because I am crap at pull requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ Product.counter_culture_fix_counts :only => [[:subcategory, :category]]
 # :except and :only also accept arrays
 ```
 
-The ```counter_culture_fix_counts``` counts method uses batch processing of records to keep the memory consumption low. The default batch size is 1000 but is configurable like so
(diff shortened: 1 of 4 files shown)


magnusvk:
Something went wrong with the diff here; this has conflicts.

magnusvk:
Instead of making these new methods, shouldn't we just make that method private? What's the point of having a public method for this in the first place?

adzap:
The methods need to be public to able to call them from the AR extension, and are the logical public interface to a counter.

Separate methods with a clearer and constrained use is optimal, as you are only doing one or the other. Calling the old update method directly could have had accidental increment without the option passed to control the operation. Things like that irk me.

Where is the conflict?

magnusvk:
screenshot 2016-05-09 10 00 10
Merge conflicts.

magnusvk:
Closing this for lack of activity.

PR #128, opened 2016-04-24
Use collection class for counter collection operations

adzap:
Moving the fix counts stuff to a collection class and then later the increment/decrement counter loops, if you merge #127. Then all significant counter behaviour is in a specific counter culture object and not in AR. 

You could then test all the CC behaviour without requiring specific individual AR models for each scenario. Only those for actually testing the AR event hooks and transaction wrapper would a model be required. But I realise that is not necessarily a goal you had in mind.

### DIFF
--- a/lib/counter_culture/collection.rb
+++ b/lib/counter_culture/collection.rb
@@ -1,0 +1,72 @@
+module CounterCulture
+  class Collection
+    include Enumerable
+
+    def initialize(arr=Array.new)
+      @arr = arr
+    end
+
+    def fix_counts(options={})
+      options[:exclude] = [options[:exclude]] if options[:exclude] && !options[:exclude].is_a?(Enumerable)
+      options[:exclude] = options[:exclude].try(:map) {|x| x.is_a?(Enumerable) ? x : [x] }
+      options[:only] = [options[:only]] if options[:only] && !options[:only].is_a?(Enumerable)
+      options[:only] = options[:only].try(:map) {|x| x.is_a?(Enumerable) ? x : [x] }
+
+      flat_map do |counter|
+        next if options[:exclude] && options[:exclude].include?(counter.relation)
+        next if options[:only] && !options[:only].include?(counter.relation)
+
+        reconciler = CounterCulture::Reconciler.new(counter, options.slice(:skip_unsupported))
+        reconciler.reconcile!
+        reconciler.changes
+      end.compact
+    end
+
+    def increment_counters(obj)
+      each { |counter| counter.change_counter_cache(obj, :increment => true) }
+    end
+
+    def decrement_counters(obj)
+      each { |counter| counter.change_counter_cache(obj, :increment => false) }
+    end
+
+    def update_counters(obj)
+      each { |counter| 
+        # figure out whether the applicable counter cache changed (this can happen
+        # with dynamic column names)
+        counter_cache_name_was = counter.counter_cache_name_for(counter.previous_model(obj))
+        counter_cache_name = counter.counter_cache_name_for(obj)
+
+        if obj.send("#{counter.first_level_relation_foreign_key}_changed?") ||
+          (counter.delta_column && obj.send("#{counter.delta_column}_changed?")) ||
+          counter_cache_name != counter_cache_name_was
+
+          # increment the counter cache of the new value
+          counter.change_counter_cache(obj, :increment => true, :counter_column => counter_cache_name)
+          # decrement the counter cache of the old value
+          counter.change_counter_cache(obj, :increment => false, :was => true, :counter_column => counter_cache_name_was)
+        end
+      }
+    end
+
+    def each(&block)
+      @arr.each(&block) if block_given?
+    end
+
+    def <<(v)
+      @arr << v
+    end
+
+    def +(y)
+      raise unless y.is_a?(self.class)
+
+      self.class.new(@arr + y.values)
+    end
+
+    protected
+
+    def values
+      @arr
+    end
+  end
+end
--- a/lib/counter_culture/extensions.rb
+++ b/lib/counter_culture/extensions.rb
@@ -5,7 +5,7 @@ module CounterCulture
     module ClassMethods
       # this holds all configuration data
       def after_commit_counter_cache
-        config = @after_commit_counter_cache || []
+        config = @after_commit_counter_cache || Collection.new
         if superclass.respond_to?(:after_commit_counter_cache) && superclass.after_commit_counter_cache
           config = superclass.after_commit_counter_cache + config
         end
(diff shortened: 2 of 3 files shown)


lib/counter_culture/extensions.rb
```diff
@@ -33,6 +33,7 @@ def counter_culture(relation, options = {})
         # add the counter to our collection
         @after_commit_counter_cache << Counter.new(self, relation, options)
       end
+      alias_method :cache_count_on, :counter_culture
```
magnusvk: What does this do?

lib/counter_culture/collection.rb
```diff
@@ -0,0 +1,45 @@
...
+        next if options[:exclude] && options[:exclude].include?(counter.relation)
+        next if options[:only] && !options[:only].include?(counter.relation)
+
+        reconciler = CounterCulture::Reconciler.new(counter, options.slice(:skip_unsupported))
+        reconciler.reconcile!
+        reconciler.changes
+      end.compact
+    end
+
+    def each(&block)
```
magnusvk: Any reason we can't just delegate these methods to the array?

magnusvk:
I'm open to moving the remaining logic in `counter_culture_fix_counts` to another place. But I don't really see that we gain a lot by making this our own custom `Collection` class. It's simpler to just keep it as an `Array`.

Why don't we just move this remaining logic into the `Reconciler`? That seems to me where it logically belongs.

lib/counter_culture/extensions.rb
adzap: sorry I missed that. That was my own semantic renaming of the method. I was leaving that for a latter PR. I will fix the PR.

lib/counter_culture/collection.rb
adzap: instead of using enumerable? Certainly could but I am also exposing the enumerable methods publicly for consistency of a collection concept.

adzap:
You could move it the reconciler but that would change its interface to a single counter to a counter collection concern in someway. But I was also going to move the array loops for the increment methods in the AR hooks into this collection class also.

I like custom collection classes. They give a tidy and powerful interface which is consistent and customisable.

adzap:
I've updated now and moved all collection operations to the collection. Now all counter operations are outside or AR.

lib/counter_culture/extensions.rb
adzap: Fixed this now.

magnusvk:
Thanks for the update. Having had some time to think about this more, I still don't think this logic belongs in a collection class. If you look at the `Collection` every single method there is either scaffolding for the custom selection or iterates over every single element. What that tells me is that the logic belongs in the individual element, not the collection.

I don't see a big gain in clarity here and so I don't think it's worth the extra complexity of a custom `Collection` class.

adzap:
These methods iterate and filter the collection of counters, that is what collection methods do. The fact that they execute something on one or many of the collection members is really the point. You can't move this logic to the member because they are essentially applied to the whole or partial collection.

Think of AR itself. It has custom collection behaviour for records e.g. pluck, destroy_all etc., which iterate the collection and run behaviour on every member. That collection is its own class(es). 

The logic which the member object needs it already has. The goal was to move the logic out of AR and into something the encapsulates the specific collection behaviour for objects controlled by the gem.

I don't see this as complex but just decomposition. The wins come with utility (I can use the counter collection behaviour independent of AR), clarity and easier changes and features later.

magnusvk:
Going to leave some more comments to be more specific about what I mean.

lib/counter_culture/collection.rb
```diff
@@ -0,0 +1,72 @@
+module CounterCulture
+  class Collection
+    include Enumerable
+
+    def initialize(arr=Array.new)
+      @arr = arr
+    end
+
+    def fix_counts(options={})
```
magnusvk: This logic is logically not related to the concept of the collection, but to the [`Reconciler`](https://github.com/magnusvk/counter_culture/blob/master/lib/counter_culture/reconciler.rb). It belongs there.

lib/counter_culture/collection.rb
```diff
@@ -0,0 +1,72 @@
...
+      each { |counter| counter.change_counter_cache(obj, :increment => true) }
+    end
+
+    def decrement_counters(obj)
+      each { |counter| counter.change_counter_cache(obj, :increment => false) }
+    end
+
+    def update_counters(obj)
+      each { |counter| 
+        # figure out whether the applicable counter cache changed (this can happen
```
magnusvk: Again, this logic has nothing to do with the collection and makes sense as executed on a single object. It belongs in the [`Counter`](https://github.com/magnusvk/counter_culture/blob/master/lib/counter_culture/counter.rb) instead.

magnusvk:
With those two chunks removed from the `Collection`, there is really nothing left there but boilerplate code to support `Enumerable`. I don't see why we should add that, then.

To be clear: I'm in favor of moving this out from the `ActiveRecord` model. I just want it to live in different places.

lib/counter_culture/collection.rb
adzap: It applies to a collection of counters. You want to perform an operation on all members of that collection. That is a valid collection method. This is a custom type specific collection. I'm not sure what you rules are about collections.

lib/counter_culture/collection.rb
adzap: counter is an instance. Collection is a collection of counters. Operations on the collection of counters can happily belong here. Otherwise you need class methods on Counter and Reconciler, which are not themselves collection classes and class methods are a crutch to generally be avoid where possible.

lib/counter_culture/collection.rb
magnusvk: I don't have any rigid rules regarding this. I just look at the problem at hand and try and find the best solution. I continue to think the domain of this code is closer to the concept of the `Reconciler` than to that of a `Collection`. So I believe it belongs there.

lib/counter_culture/collection.rb
magnusvk: I don't follow a rule to avoid class methods whenever possible. Also, iterating over a collection is an easy enough operation that you could push it to the caller if you wanted to.

magnusvk:
@adzap I really appreciate your contribution here but I just don't think adding the concept of a `Collection` here improves the codebase. I'd be happy to accept a pull request that moves this logic into the `Reconciler` and `Counter` instead if you're willing to work on that. If not, we might just have to agree to disagree.

Thanks again for working on this!

djsegal:
@magnusvk I don't know if this is related, but my test suite breaks because `destroy_all` doesn't update the count:

i.e.

```
expect{ user.comments.destroy_all }.to \
  change{ user.comments_count }.from(N).to(0)
```

magnusvk:
@djsegal I don't think it's related. Do you use transactional fixtures? That will not work as counter_culture updates counts after commit. This is [documented in the Readme](https://github.com/magnusvk/counter_culture/#a-note-on-testing).

There is really no need for this test in your test suite. You are just duplicating existing counter_culture tests.

magnusvk:
Going to close this as it's been a while since there was activity on this PR.

PR #131, opened 2016-05-17
Added custom lambda support to :touch argument

DavidMikeSimon:
This lets a user be more specific about how their cache invalidation fields work. For example, I needed this feature because my project uses random UUIDs rather than timestamps for cache expiration.


magnusvk:
I'm not sure this is a common-enough use-case to include this extra complexity in the gem for everybody.

Might we be able to solve your use-case by firing `after_touch` callbacks on the record? Then you could hook into that.

DavidMikeSimon:
That works provided that :touch won't complain about the lack of timestamp fields in my model. I'll look into that tomorrow.

PR #133, opened 2016-05-23
wrap fix_counts in transactions

chrisdevor:
We had to update over 10 mil records.  Wrapping each batch in #counter_culture_fix_counts in a transaction reduces each update query time from about 8.0ms to 0.3ms.  In total, saved us waiting >24 hours for the update to complete.

before:
image

after:
image


olivierlacan:
@magnusvk could you please merge this? I've had lots of issues running fix counts on multi-million record tables, this would help a lot. 

❤️

magnusvk:
Thanks for the push; I wanted to do some research about this before merging but then forgot about it. Seems like this really will be faster. Merged and released as gem version 0.2.3.

Thanks for contributing this @chrisdevor and sorry for the long wait.

PR #135, opened 2016-06-01
support delta_magnitude for dynamic update value

Madao-3:
If we have scenario like : 
`  counter_culture :image, column_name: :weight, delta_magnitude: Proc.new { daynamic value } `

then we can define delta_magnitude with a Proc or number, not just 1

### DIFF
--- a/spec/counter_culture_spec.rb
+++ b/spec/counter_culture_spec.rb
@@ -663,6 +663,23 @@ describe "CounterCulture" do
     industry2.rexiews_count.should == 1
   end
 
+  it "increments dynamic delta magnitude" do
+    user = User.create
+
+    user.review_value_sum.should == 0
+
+    twitter_review_using = TwitterReview.create :user_id => user.id, :review_type => 'using'
+    user.reload
+
+    user.review_value_sum.should == 2
+
+
+    twitter_review_using = TwitterReview.create :user_id => user.id
+    user.reload
+    
+    user.review_value_sum.should == 1    
+  end
+
   it "increments dynamic counter cache on create" do
     user = User.create
     product = Product.create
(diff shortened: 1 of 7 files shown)


magnusvk:
Thanks for this. This needs a test, though. Could you add one?

Madao-3:
ok, the test is finish.

and there is a tiny problem about proc usage.

https://github.com/JuanitoFatas/ruby-style-guide#proc

i think it will make the code look better. ;-)

spec/counter_culture_spec.rb
```diff
@@ -663,6 +663,23 @@
...
+    twitter_review_using = TwitterReview.create :user_id => user.id, :review_type => 'using'
+    user.reload
+
+    user.review_value_sum.should == 2
+
+
+    twitter_review_using = TwitterReview.create :user_id => user.id
+    user.reload
+    
+    user.review_value_sum.should == 1
```
magnusvk: Shouldn't it be 3 now?

magnusvk: Can you also destroy the `TwitterReview`s and test that the decrement works correctly?

magnusvk:
Unfortunately the test fails in CI. Can you look into what's going on?

Thanks!

Madao-3:
I'm so sorry about that. Now it have passed.

Madao-3:
@magnusvk  : )

magnusvk:
I had to tweak / add a couple of things:
* Added support for fix_counts
* Added a test to destroy
* Added a test for static delta magnitude
* Tweak README

To get this ready to merge. But I kind of wanted this in the gem, so I just went ahead and did those. You can see these changes at 950c3f3e34f0c3f2fb1eed35b1322f87ef7a2971.

That's now released as gem version 0.2.1.

Madao-3:
cool 👍

PR #137, opened 2016-06-15
Updated homepage link in gemspec

andrew:
Thanks for the gem, I'm using it in https://libraries.io 🍻


magnusvk:
Thank you. Kind of hilarious this was still lurking here.

Issue #30, opened 2013-11-18
Manually filling in dynamic column name based on joined model


dzoltok:
OK, so I have a situation where an Article has_many Tags, and a Tag has_many Articles, through a join table that belongs_to both of them. That's where this code is:

```
class ArticlesTag < ActiveRecord::Base
  belongs_to :article
  belongs_to :tag

  counter_culture :tag, column_name: "articles_count"
  counter_culture :tag, column_name: Proc.new{ |model| model.article.recent? ? 'recent_articles_count' : nil }
```

As you can see, I'm trying to store both the number of articles associated with a tag, and the number of "recent" articles. The problem here comes up when I try to populate the count columns in the Tag table using counter_culture_fix_counts:

```
Must provide :column_names option for relation [:tag] when :column_name is a Proc; you may skip this relation with :skip_unsupported => true
```

But I can't figure out how to properly configure that option. The condition that makes an article recent is on the article table itself, not the join table where all the counter_cache code is going. If I add an option like:

```
column_names: {
  ["articles.created_at > ?", 30.days.ago] => 'recent_articles_count'
}
```

I just get an error that there is no such column 'articles.created_at', since the articles table isn't being joined in when it tries to run the counter_culture_fix_counts command.

swapnilabnave:
I would like to [contribute](https://github.com/magnusvk/counter_culture#contributing-to-counter_culture) and add an option for `joins`. Would that be a good thought ?
I know it wont be efficient in terms of space and time but it would be an **option** eventually.

```ruby
counter_culture :tag,
   column_name: Proc.new{ |model| model.article.recent? ? 'recent_articles_count' : nil },
   joins: ["articles"],
   column_names: { ["articles.recent = ?", true] => 'recent_articles_count' }
```

magnusvk:
@swapnilabnave I'm not sure I understand how that would work. I'm open to adding some sort of support here going forward, but I'd need some more details to know whether that makes sense.

swapnilabnave:
@magnusvk 
Ok. So here's real case scenario, which I was trying to achieve using [`counter_culture`](https://github.com/magnusvk/counter_culture).

```ruby
class Conversation < ActiveRecord::Base
  attr_accessible :name, :guest_count

  has_many   :conversation_members
  has_many   :guest_users, through: :conversation_members, source: :user, conditions: ["users.user_type = ?", 'G']
end

class ConversationMember < ActiveRecord::Base
[...]

PR #138, opened 2016-06-25
Fixes: #30 | Add support for joining tables while fixing_counts(Model#counter_culture_fix_counts)

swapab:
@magnusvk That's pretty much I had thought about.

I feel the `CounterCulture::Reconciler` class can be further abstracted out to be more `CounterCulture::Join` specific. Thoughts?


swapnilabnave:
Any word on this one ?

magnusvk:
Sorry, was AFK for a bit. Just getting back to this.

Thinking about this some more: Is there a way to make this automatic instead? It rankles me a bit that one still has to specify two options here. What if `column_names` supported a hash for this. As an example:
```ruby
:column_names => {
    ["products.product_type = ?", 'awesome'] => 'awesome_count',
    ["products.product_type = ?", 'sucky'] => 'sucky_count'
```
would become
```ruby
:column_names => {
    [{ products: :product_type }, 'awesome'] => 'awesome_count',
    [{ products: :product_type }, 'sucky'] => 'sucky_count'
```

This looks like a better way to pass these options anyhow, looking at this today. And it would allow us to automatically infer if another join is necessary here?

Or am I overthinking this?

swapnilabnave:
Hash looks better. I'll make the change soon.

magnusvk:
@swapnilabnave Any news on this?

swapnilabnave:
@magnusvk It went out of my radius.
I'll update the patch soon.

swapnilabnave:
@magnusvk 
I actually have a question.

If we are allowing the hash structure of `column_names` option, then the SQL conditions such as `IN` won't be useful.

For example: following query can't be defined or recognized in hash syntax.

`"products.product_type IN (?)", ['awesome', 'superb']`

```ruby
:column_names => {
    [{ products: :product_type }, 'awesome'] => 'awesome_count',
    [{ products: :product_type }, 'sucky'] => 'sucky_count'
```
--
To keep the current syntax and make the joins 'automatic', can we instantiate the `@joins` variable based on the conditions those are being defined on `column_names`. So if the conditional table is different then the applied model's table then we join that table else we don't.

_I can code this if my words sounds confusing._

swapnilabnave:
@magnusvk 
This is complete from my side. Can you please review ?

magnusvk:
Hm; we can do the `IN` automatically based on the object type, right? If it's an `Array` (or `Enumerable`, really) then we use `IN`, if not we use `=`. Just like how `where()` works in ActiveRecord. Am I missing something?

krim:
Hello. I need "joins" functional too. What's wrong with this PR?

shaneog:
I think @swapab needs to rebase this to fix the conflicts for @magnusvk to consider it.

magnusvk:
Yeah unfortunately looks like this PR may be abandoned at this point...

magnusvk:
Closing this, feel free to reopen if there's progress on this work.

olivierlacan:
Aside from a rebase, what would you want for this PR to be complete @magnusvk?

magnusvk:
See this comment for changes I was still hoping for here:
https://github.com/magnusvk/counter_culture/pull/138#issuecomment-231856739

I'd still be willing to consider this change or one like it if you wanted to pick up the work.

PR #139, opened 2016-06-29
Use `ActiveSupport.on_load` to hook into ActiveRecord

rono23:
There is a weird behaviour of config `belongs_to_required_by_default` on Rails 5.0.0.rc2 when gem is not playing nice with the rails hooks.

Related rails/rails/issues/23589


magnusvk:
Thank you! Released as gem version 0.2.2 with this change.

PR #142, opened 2016-08-27
add required components of activesupport

namusyaka:
Hi.
I'd like to use this gem in my padrino project. Padrino hasn't depend on activesupport now, so these components should be `require`d .

What do you think?


magnusvk:
Sounds good! Thanks for this.

Issue #120, opened 2015-12-22
Move counter cache update back into transaction


magnusvk:
Currently, the counter cache update happens outside of the transaction in the after_commit hook. This is to avoid a historical deadlock issue in Postgres. However, this should be fixed in Postgres 9.3, [according to this](http://mina.naguib.ca/blog/2010/11/22/postgresql-foreign-key-deadlocks.html) and the release notes. 

Given that, it should be optional in counter_culture whether the update should happen in or outside of the transaction. It should default to inside the transaction.

magnusvk:
cc @DavidMikeSimon

c0m3tx:
Hi all, any news about this?

magnusvk:
No news, no. Feel free to work on it though! 😄 I've never worked with the test_after_commit gem, so no ideas there. How are you trying to use it?

magnusvk:
I think it's even simpler than that. It'd just need a method that conditionally yields directly or calls `execute_after_commit`.

kulte:
@magnusvk Check out #145. Let me know if you'd like me to add any specs, but I'm not really sure how to reliably test what happened inside and outside of an ActiveRecord transaction.

magnusvk:
This is no longer the case in the just-released version 1.0.0. Thank you @kulte for the pull request.

jbritten:
I'm experiencing deadlocks using MySQL & counter_cache at scale (database under heavy load with many concurrent requests competing to update the counters).

Any chance the "execute_after_commit" option can be added back? I don't understand why it was removed from 2.x in the first place. Seems that whether to update the counters inside or outside of the transaction should be left to the user to decide, right?

I cannot revert to the 1.x branch because I need other improvements in the 2.x branch (i.e. bug fixes for discard integration).

magnusvk:
Sorry you're having issues with this. Interestingly, this issue was first added because of deadlock issues with Postgres. It's somewhat ironic that it's now MySQL causing a similar issue.

In any case, the history of removing this comes from a bugfix to allow multiple saves in one transaction, see 81dfbf55495b59afab844bbd0ff14fbff3cb922b. Plus, in the absence of issues like this with the database layer, I don't see why you'd ever want to push the counter cache update outside of the transaction. But then again, you _are_ seeing database issues.

[...]

PR #145, opened 2016-11-10
Fixes: #120 | Move counter cache update back into transaction by default

kulte:
Make it optional in `counter_culture` whether the update should happen in or outside of the transaction. It should default to inside the transaction.

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -1,6 +1,6 @@
 module CounterCulture
   class Counter
-    CONFIG_OPTIONS = [ :column_names, :counter_cache_name, :delta_column, :foreign_key_values, :touch, :delta_magnitude ]
+    CONFIG_OPTIONS = [ :column_names, :counter_cache_name, :delta_column, :foreign_key_values, :touch, :delta_magnitude, :execute_after_commit ]
 
     attr_reader :model, :relation, *CONFIG_OPTIONS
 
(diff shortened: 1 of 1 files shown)


lib/counter_culture/counter.rb
```diff
@@ -14,6 +14,7 @@ def initialize(model, relation, options)
       @foreign_key_values = options[:foreign_key_values]
       @touch = options.fetch(:touch, false)
       @delta_magnitude = options[:delta_magnitude] || 1
+      @execute_after_commit = options.fetch(:execute_after_commit, false)
```
mrclmrvn: @kulte, should this be `true` by default when not set, as it's the current behaviour?

I agree, this would be a useful option.

kulte: @mrclmrvn I agree, although if you look at @magnusvk original request in #120, he had wanted it to default to be in the transaction. IMO this would be a breaking change and require a major version bump, but as a non-maintainer, that falls out of the scope of a PR per the README.

Happy to default to `true` though if that's what the maintainer(s) want.

mrclmrvn: Got it!

magnusvk:
@kulte Thank you for this. You are right, I think the default should be false given that there isn't really a reason to do this in any database as long as you're running a recent version. Releasing a 1.0.0 was way overdue anyhow I would say. 

I've updated the Readme and Changelog and released version 1.0.0 to Rubygems just now.

mrclmrvn:
Great news!

PR #149, opened 2016-12-14
Create spec for issue#148

kmcgaire:
So it was an error on our end. However if you want to keep the extra specs feel free!


magnusvk:
Thank you!

PR #150, opened 2016-12-29
Reduce counts of batch-testing specs (except in CI)

timdiggins:
Have started working on some changes to counter culture. It's frustrating to have a test run with such long pauses. I'm not 100% sure why such large numbers are needed, but I think the large numbers could only happen in CI where the long pause won't be noticed.

total test run before PR: ~ 30s
total test run after PR: ~ 5s


magnusvk:
Yeah; the idea is to test that the batch processing works correctly. Probably doesn't need these big numbers at all but this looks like a good idea. Thank you.

PR #152, opened 2016-12-29
Show all failure on ci test run

timdiggins:
Show all failures when running on CI test run

requires/includes #150


magnusvk:
Thank you.

PR #153, opened 2016-12-29
Update rspec to latest

### LINKED ISSUE #151: Update to latest rspec (3.x)?
I'm starting to work on a very basic very of polymorphic associations (#4) but would like to use more recent version of rspec style. I'd like to update rspec to the latest v2 which allows for both rspec styles, but really wouldn't be much bother to convert it wholesale to v3 (using transpec). I'll start working on a parallel PR for this, but do let me know whether you'd appreciate it, or if there's a reason to stay with old-style rspec (`.should`).

timdiggins:
Also modernizes specs from `.should` style to `expect()` style.

Requires/includes #150 #152 and fixes #151


magnusvk:
Thanks for doing this; I've just been too lazy to actually do this. The new syntax is so much better.

PR #154, opened 2016-12-29
Polymorphic association (single-level only)

timdiggins:
Just checking in to show you the work in progress on polymorphic associations support. 

Aim is to just support  polymorphic associations on single-level relations only (ie. counter_culture relation length is 1/a symbol).


timdiggins:
Custom column names and conditional counts now tested

timdiggins:
Hey @magnusvk, just pinging  to check you've seen this and the other PRs. Have more refactoring I could do but want to check you're up for it first.

lib/counter_culture/reconciler.rb
```diff
@@ -19,123 +19,161 @@ def reconcile!
...
         raise "Must provide :column_names option for relation #{relation.inspect} when :column_name is a Proc; you may skip this relation with :skip_unsupported => true" if counter_cache_name.is_a?(Proc) && !column_names
         raise "Fixing counter caches is not supported when :delta_magnitude is a Proc; you may skip this relation with :skip_unsupported => true" if delta_magnitude.is_a?(Proc)
+        raise "Fixing counter caches is not supported on polymorphic associations unless you specify `polymorphic_associated_models: ['Model1', 'Model2']` on the counter_cache" if polymorphic? && polymorphic_associated_models.nil?
       end
 
-      # if we're provided a custom set of column names with conditions, use them; just use the
-      # column name otherwise
-      # which class does this relation ultimately point to? that's where we have to start
+      if polymorphic?
+        polymorphic_associated_models.each do |model|
```
magnusvk: Couldn't we just do a `.pluck("DISTINCT #{relation_name}_type")` to get the models here?

lib/counter_culture/counter.rb
```diff
@@ -176,6 +201,12 @@ def first_level_relation_foreign_key
       relation_reflect(first_relation).foreign_key
     end
 
+    def polymorphic?
+      relation_reflect(relation).polymorphic?.tap do |is_polymorphic|
```
magnusvk: I find this use of tap to be confusing, frankly. `</nitpick>`

lib/counter_culture/counter.rb
```diff
@@ -176,6 +201,12 @@ def first_level_relation_foreign_key
...
+        raise "Polymorphic associations only supported with one level" unless (relation.is_a?(Symbol) || relation.length == 1) if is_polymorphic
```
magnusvk: Is this a necessary constraint? Or have you just not tested that yet?

lib/counter_culture/reconciler.rb
timdiggins: Ok I hadn't thought of that. I'll see if it works throughout and we can ditch that option.

lib/counter_culture/counter.rb
timdiggins: I thought this was a necessary constraint -- but I never fully got my head around multi-levels. But let me think it through here:

You've got a 

```ruby
class Category; end
```

and
```ruby
class Employee; belongs_to :company; has_many :images, as: :imageable; end
```

and

```ruby
class Building; belongs_to :company; has_many :images, as: :imageable; end
```

and then a 

```ruby
class Image
  belongs_to :imageable, polymorphic: true
  # easy:
  counter_culture :imageable
  # possible (but why)
  counter_culture [:category, :imageable]
  # more likely, but still seems contrived
    counter_culture [:category, :imageable], :column_name => proc {|model| "#{model.imageable_type.downcase}_image_count" }

end
```

Ok -- that seems reasonable enough, as a test case. Now I guess I could try implementing it (might just work... already). If it gets tricky, I might just back out and leave this to someone who actually needs it.

timdiggins:
@magnusvk The couple of refactorings I wanted to do (but I haven't got a strong feel on exactly how):

1)I felt something in the Counter class a bit confusing -- that it both has a `relation` but also many of the methods also take a `relation`. The mixture of utility methods (should be class methods? or maybe a different class?) with ones that are really messages to this class felt like a confusion.

2) I felt like there was some kind of duplication around how the relation itself was managed. it would be nice to just resolve it once (in the initializer maybe) rather than always having to take a symbol or an enumerable -- or maybe just make this symbol-or-enumerable a  class in its own right. But I think that improving 1 above would make the path for this one clearer. (or vice versa...)

Would be useful to know if you'd find either of these welcome (if successful) or if you'd already had thoughts on these lines.

magnusvk:
Yeah those both make sense; I've been meaning to do 1) myself and just haven't gotten to it. So would love those changes; probably best in a separate PR if possible though.

By way of historical context: counter_culture used to be kind of spaghetti-code-ish in that everything was in one big file that'd get mixed into `ActiveRecord::Base`. Then a kind soul started refactoring and pulling stuff out which is the current state; but some of the code still assumes the old organization (like having to pass in the `relation`). So, no attachment to that way of doing it.

lib/counter_culture/reconciler.rb
timdiggins: Have done this -- but using rails api (AssociationReflection#foreign_type) just in case of overriding default

lib/counter_culture/counter.rb
timdiggins: done

lib/counter_culture/counter.rb
timdiggins: @magnusvk I've started on this, but realize there's more than one way of having a polymorphic association in a multi-level relation (beginning, middle and end). I'd like to work on support for **multi-level** polymorphic as a separate PR (maybe after the other two refactoring PRs) -- would you be happy to merge this one as is?

timdiggins:
@magnusvk can you consider merging this for single level and I'll look at multi level in a separate PR

magnusvk:
@timdiggins Definitely open to that; I just haven't had time to look over this yet. Will let you know when I do. Thanks again for working on this!

magnusvk:
@timdiggins Sorry for the long delay here!

I pushed commits with some nit-picky changes and to add Rails 3.2 compatibility. I also merged in the latest master.

However, I stumbled on one problem: I also added some more tests. You were only testing the counter cache update on create, not on destroy or update. I added tests for those cases; unfortunately this doesn't seem to work in the polymorphic case.

Can you take a look at what's going on with that?

Thanks for your contributions here!

timdiggins:
good catch on the update @magnusvk and thanks for bringing up to date with master. Will have a look and see what I can work out

timdiggins:
@magnusvk have now fixed that spec and simplified a bit (travis ci is reporting a fail in the PR build but it's a gem install timeout -- think it will be fine if you kick off a rebuild)

magnusvk:
@timdiggins thank you. I merged in master again and cleaned this up a bit. Want to look over my changes? I think this looks good to go now from my point of view.

timdiggins:
@magnusvk hooray! Have looked through changes -- LGTM.

On a side note, I will try to find time to have a look at refactoring a bit -- take out the methods that take a relation out of Counter and into a Utility module, and have the Counter just use own relation, and maybe look at having a class that represents a counter plus its source / obj (the pair is used a lot).

Also might try to see if I can make a spec for multi-level polymorphic associations and make them work (just to be able to remove the warnings).

Another thought I had (because I had to do it in a repo using counter-culture) is to have an utility method available from counter culture to "fix counts slowly" which takes just for a single instance or scope (the target rather than source of the counter culture) -- which means you can just use the procs for delta/column_name without rewriting as sql (won't scale or work in many cases, but can be useful) 

Just ideas. (PS apologies for hijacking the PR thread for this random chat 😨 )

magnusvk:
@timdiggins all of these sound like good ideas. 😄 Happy to look at those PRs if / when you have something to look at! And thanks again for making this happen.

PR #157, opened 2017-01-18
remove rvm-specific files from repo.

timdiggins:
It is a real pain to have rvm files (.ruby-version) in the repo.
We already define which ruby versions are supported in travis.yml 
and I think it should be up to developer which version to develop on,
so these should be local files.

This will be a slight pain for existing developers who are using rvm,
who will need to recreate the .ruby-version locally.


magnusvk:
Hah. Sure.

PR #161, opened 2017-02-18
support custom touch column

madmax:
We should be able (like in Rails) touch custom columns when we specify column name instead true  for touch. 

    counter_culture :product, touch: :rexiews_updated_at


magnusvk:
Thank you! Released as 1.2.0 just now.

PR #164, opened 2017-03-09
Use ruby >= 1.9 hash syntax

dsakuma:
Hi! 

This PR is just updating the documentation with new hash syntax introduced by ruby 1.9. I'm using counter culture with this new syntax in my project and everything is working fine.


magnusvk:
Thank you. Why not! :) It should work with either syntax.

PR #165, opened 2017-03-20
reduce noise in test runs when running locally

timdiggins:
stops the local migrations from printing out


magnusvk:
This seems reasonable. Any reason we wouldn't want this to also apply in CI, though? Personally I'd prefer to keep it consistent.

timdiggins:
Agreed. Have tweaked commit.

magnusvk:
Thank you!

PR #172, opened 2017-05-26
Allow updating of specific instance to trigger Papertrail versioning.

key88sf:
@magnusvk This isn't super pretty, but it seems to work. Basically if you have a model where you need to also record Papertrail versioning for the counter column, you can specify the `with_papertrail` option and it will load the specific instance of the object being modified (with row-level locking), do the counter update, and record a papertrail version row.

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -15,6 +15,7 @@ module CounterCulture
       @touch = options.fetch(:touch, false)
       @delta_magnitude = options[:delta_magnitude] || 1
       @execute_after_commit = options.fetch(:execute_after_commit, false)
+      @with_papertrail = options.fetch(:with_papertrail, false)
     end
 
     # increments or decrements a counter cache
@@ -28,6 +29,7 @@ module CounterCulture
     #   :was => whether to get the current value or the old value of the
     #      first part of the relation
     #   :execute_after_commit => execute the column update outside of the transaction to avoid deadlocks
+    #   :with_papertrail => update the column via Papertrail touch_with_version method
     def change_counter_cache(obj, options)
       change_counter_column = options.fetch(:counter_column) { counter_cache_name_for(obj) }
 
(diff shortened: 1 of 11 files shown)


key88sf:
FWIW - updated to just trigger a version row to be created. In this case the `object_changes` part won't reflect the change, but the values in the `object` stored by papertrail will be the updated values. This should typically be good enough for someone doing analysis on this data.

key88sf:
@magnusvk If you don't think this belongs in the main gem, let me know and we can close the PR and just use a fork.

magnusvk:
@key88sf doing this for just Papertrail seems awfully specific. I remember you having a different version of this where you did this with a full-on `save` to trigger callbacks? That seems more appropriate to me for inclusion in master; then the option would be `with_callbacks` which is just much more generically useful.

What do you think?

key88sf:
@magnusvk Yeah that's fine though it results in less performant code (has to reload the object, and also will trigger other on_save callbacks as well).  This is why we decided to just touch a row specific to papertrail. For our needs, the single papertrail row was all we ended up needing, though others might find the more generic solution more beneficial.

If you'd like I can re-open a PR based on that original generic solution. WDYT about supporting both??

magnusvk:
Yeah that's a good point. Looking at this again, I'm ok with merging this even without the additional save support. Another advantage of this PR is that it's less invasive to how updates are done.

If you want to work on supporting a full `save` that's something that'd be great. But I'm happy to merge this, too.

But! It does need a test. Mind adding one?

Thanks for contributing!

lib/counter_culture/counter.rb
```diff
@@ -42,6 +44,10 @@ def change_counter_cache(obj, options)
                           else
                             counter_delta_magnitude_for(obj)
                           end
+
+        klass = relation_klass(relation, source: obj, was: options[:was])
+        primary_key = relation_primary_key(relation, source: obj, was: options[:was])
+
```
key88sf: Hmm yeah I think this was leftover from the original change. Can revert this part.

key88sf:
@magnusvk Ok added spec coverage :)

magnusvk:
@key88sf I tweaked this a bit, added documentation and released as 1.7.0 just now. Thanks for contributing!

key88sf:
👍

Issue #177, opened 2017-07-05
Conditional counts not updating as expected


CodingAnarchy:
I am working with a `Child` class with an `active` boolean field, and this conditional count to reference another model:

```Ruby
column_name: proc { |model| model.active? ? 'children_count' : nil }
```

When adding one of these child classes to another model with the counter, the `active` flag is not being appropriately handled.

The `<<` methodology will always add to the counter in memory, but will be rejected upon reloading the instance, regardless of whether `child.active?` is `true` or `false`.  Additionally, `children.delete(child)` on the association will always reduce the count, regardless of the `active?` check, while `child.delete` has not adjusted the count at all.

I would expect that only the `active` child instance would produce an increase or decrease in the counts as expected.  If needed, I can reproduce my specs that demonstrate this odd behavior.

magnusvk:
You don't have enough information here to figure out what's going on, unfortunately. I don't even know what kind of ActiveRecord relationship you're using between these two classes.

Could you open a pull-request with a failing test for me?

CodingAnarchy:
I've opened #178 as a pull request with some failing tests.

The ActiveRecord relationship is a basic `has_many`, and I was looking to cache just the `active` count of the subordinate class.

CodingAnarchy:
I've got it working as expected now.  Thanks for the help.  I'll close this issue.

PR #178, opened 2017-07-05
Failing specs added for #177

CodingAnarchy:
I've added some specs that replicate those that I am running when I run into the issue noted with #177.

Please let me know if there is anything I can do to clarify this behavior.  (The application I duplicated these specs from is running Rails 4.2.1 presently.  This still fails on Rails 5.1, but not in the same manner on my local machine.)

### DIFF
--- a/spec/counter_culture_spec.rb
+++ b/spec/counter_culture_spec.rb
@@ -400,6 +400,68 @@ describe "CounterCulture" do
     expect(user2.tried_count).to eq(0)
   end
 
+  describe "conditional counts on addition/removal from association" do
+    let!(:main) { ConditionalMain.create }
+    let!(:condition) { ConditionalDependent.create(condition: true) }
+    let!(:non_condition) { ConditionalDependent.create(condition: false) }
+
+    it "should increase cache in database on adding a true conditional" do
+      expect {
+        main.conditional_dependents << condition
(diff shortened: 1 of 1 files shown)


magnusvk:
@CodingAnarchy I see two failures with this:
1. The in-memory checks fail. counter_culture is just not designed for this, it changes counters in the database only using an `UPDATE` statement.
2. Calling `delete` on a relation doesn't update the counter. That's because `delete` does not trigger callbacks, thus there's no way for counter_culture to know about this change. If you change it to `destroy` instead, it works correctly.

If you look at #179 you see that making those changes makes these specs go green. Am I missing something?

CodingAnarchy:
Ok, so it sounds like the in-memory stuff was a misunderstanding on my part, and I'm switching to use `destroy` as well just to ensure that the callbacks are done.

It looks like in #179, the older versions of Rails still fail the specs, though.  And looking at the rails history for updates regarding the implementation of `counter_cache`, there was a change made in 4.2 that was a fix for 4.1.9, and then another change was made for 4.2.1 to fix another issue that caused, so it looks like maybe Rails 4.2.1 just happens to be a blind spot here.  I'll try running some checks on different 4.2.X versions to see if that might be part of the issue I'm seeing as well.

CodingAnarchy:
Closing in favor of #179 which shows the real issue with the older Rails versions.

PR #179, opened 2017-07-06
Updated specs from #178

magnusvk:
This is just so we can run specs without me pushing to #178.


magnusvk:
Interestingly, the behavior here seems to depend on the Rails version you are using: https://travis-ci.org/magnusvk/counter_culture/builds/250976883?utm_source=email&utm_medium=notification

PR #188, opened 2017-08-10
Correctly quote table names and use aliases consistently

magnusvk:
So that using this with PostgreSQL and schema namespaces will work
correctly.


magnusvk:
I'd love to have a spec for this, but I can't figure out how to without adding a Postgres database to run tests against which I don't want to do. So I'm just going to merge this... :(

PR #190, opened 2017-08-24
Fix version check (should check ActiveRecord instead of Rails)

bencollab:
When using counter_culture with a non-Rails project, the Rails version conditionals are failing (because `Rails` is not defined). This PR replaces those conditionals with checks for the ActiveRecord version.

I've successfully verified that this change allows it to work with a project that relies on ActiveRecord (but not Rails).


magnusvk:
Thank you!

magnusvk:
This is included in the just-released 1.8.0 gem release.

PR #191, opened 2017-08-31
Check ActiveRecord version instead of Rails version in Reconciler

bencollab:
My apologies for missing this in my previous PR.

This adds to the changes in https://github.com/magnusvk/counter_culture/pull/190 to allow for reconciliation in non-Rails projects.

### DIFF
--- a/lib/counter_culture/reconciler.rb
+++ b/lib/counter_culture/reconciler.rb
@@ -3,6 +3,8 @@ require 'active_support/core_ext/module/attribute_accessors'
 
 module CounterCulture
   class Reconciler
+    ACTIVE_RECORD_VERSION = Gem.loaded_specs["activerecord"].version
+
     attr_reader :counter, :options, :changes
 
     delegate :model, :relation, :full_primary_key, :relation_reflect, :polymorphic?, :to => :counter
@@ -221,7 +223,7 @@ module CounterCulture
       end
 
       def parameterize(string)
-        if Rails.version < '5.0'
+        if ACTIVE_RECORD_VERSION < Gem::Version.new("5.0")
           string.parameterize('_')
         else
           string.parameterize(separator: '_')


bencollab:
As far as I can tell, the CI test failures appear to be unrelated to this change.

magnusvk:
@bencollab the failure is:
```
     NoMethodError:
       undefined method `gem_version' for ActiveRecord:Module
```

But only when run against Rails 4.0. Seems very related.

bencollab:
@magnusvk Yes it's related, but only coincidentally; as far as I can tell, it's not *caused* by this change. The exception is being raised in the paper_trail gem, specifically this line that was added 12 days ago:
https://github.com/airblade/paper_trail/blame/v7.1.2/lib/paper_trail/version_concern.rb#L15

Relevant discussion on the paper_trail gem: https://github.com/airblade/paper_trail/pull/956

Apparently the `ActiveRecord.gem_version` method wasn't added until AR v4.2, and they included a `respond_to?` conditional for the method when checking versions in another place in the paper_trail codebase, but it looks like they forgot to do something similar in `./lib/paper_trail/version_concern.rb:15`. AR v3.2 isn't triggering the exception because the latest (problematic) version of paper_trail requires AR >= v4.0, so an older version of paper_trail is used when AR v3.2 is in use.

The same exception is raised when running `rspec` on my own machine (with Rails v4.0.13) before this change (using commit 6d8f0f5) when paper_trail v7.1.2 is installed.

magnusvk:
I opened https://github.com/airblade/paper_trail/pull/988 to address this in PaperTrail.

magnusvk:
@bencollab thanks for this. I merged this via a cherry-pick and just skipped the PaperTrail tests for Rails 4.0. Released as 1.8.1.

PR #198, opened 2017-09-25
Updated gem install in readme

tibbon:


magnusvk:
@tibbon I disagree with locking in the minor version, must people should just upgrade those as they're always compatible with how I version. But, sure, I've updated this to point to the latest version in the Readme. But, inevitably, it'll end up out of date again. ;)

PR #200, opened 2017-09-27
make use of batch_size parameter

mattias-lundell:
add batch_size to whitelisted options passed to Reconciler


magnusvk:
Thanks @mattias-lundell, much appreciated. Released as 1.8.2.

PR #201, opened 2017-09-28
Add support for composite primary keys

mattias-lundell:
Change the way the column that we count over is calculated to support
composite primary keys.

NB: don't know how to test this without adding a dependency to
composiite-primary-keys gem.

Update: tests fails on older rails.


magnusvk:
Can you explain more what this does? I don't really understand without a test.

You can work around the older Rails versions failing with version checks like we already do in other places.

And it's totally fine to add a gem, just put it in the `:development, :test` section of them Gemfile. We already do this with other gems we need for testing, like Paperclip.

magnusvk:
Closing this for lack of update.

PR #203, opened 2017-11-20
Change  hash rocket to symbol at template migration file

kyohei0423:
I suggest using symbol at template migration file.


magnusvk:
Not sure why you closed this, but I agree. Did this in https://github.com/magnusvk/counter_culture/commit/91f9e000cac2a3b61f902bb5ada64f5b3413c189

PR #204, opened 2017-11-25
Error when default_scope is used at mid-level

cuongdh21:
In the default_scope of second-level model, there is a join with the third-level model. The error will looks something like:
```
Column 'reviews_count' in field list is ambiguous: UPDATE `companies` INNER JOIN `indistries` ON `companies`.`id` = `industries`.`company_id` SET `reviews_count` = COALESCE(`reviews_count`, 0) + 1, updated_at = '2017-11-25 06:35:09' WHERE `companies`.`id` = 2
```
So the catch is it's trying to update the ```reviews_count``` column of the join table in which there are duplicated ```reviews_count``` column, hence ```ambiguous```.
I had a fix for this, but havent been able to replicate the issue in my test-case.

### DIFF
--- a/spec/counter_culture_spec.rb
+++ b/spec/counter_culture_spec.rb
@@ -508,6 +508,28 @@ describe "CounterCulture" do
     expect(product.reviews_count).to eq(1)
   end
 
+  it "increments third-level counter cache on create when default scope at second-level" do
+    Company.with_default_scope do
+      industry = Industry.create
+      company = Company.create :industry_id => industry.id
+      user = User.create :manages_company_id => company.id
+
+      expect(industry.reviews_count).to eq(0)
+      expect(company.reviews_count).to eq(0)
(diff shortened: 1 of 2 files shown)


magnusvk:
I'm confused by this, can you clarify what the error is? The test case you added doesn't appear to be failing.

cuongdh21:
Yeah I could not replicate the error I got in my situation in test case. In my case it was trying to update counts of a join table of second and third level models (the second level model used a default_scope that join and preload the third level model), hence ambiguous column name error for count column. However, when I created the test case, the way the query worked look different (it did not update the join table) and the test passes. I tried a bit but havent been able to figure out the discrepancy in how the queries work.

magnusvk:
Sorry about the trouble you're having, but I don't really know how to address this without a failing test case, so going to close this for now. Feel free to reopen if you can find a way to reproduce this for me.

PR #205, opened 2017-11-28
[fix] generate migration template with version for rails >= 5

howtwizer:
In Rails > 5 need to insert rails version to migrations. 
This fix inserting this version only for Rails versions > 5

Caused by:
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class AddMessagesCountToConversationies < ActiveRecord::Migration[4.2]


magnusvk:
Thank you!

magnusvk:
Released 1.9.0 with this.

PR #207, opened 2017-12-07
improvement: avoid extra query in case of unchanged relation

adone:
I have many `counter_culture` in one model with dynamic counter cache columns
and I noticed `counter_culture` does many queries for relation even it has already loaded and has not been changed
for example
```
    SQL (1.2ms)  UPDATE "votes" SET "value" = $1, "updated_at" = $2 WHERE "votes"."id" = $3  [["value", 0], ["updated_at", "2017-12-07 09:16:36.723428"], ["id", 16]]
  Post Load (1.2ms)  SELECT  "posts".* FROM "posts" WHERE (posts.id = 64518) ORDER BY "posts"."id" ASC LIMIT $1  [["LIMIT", 1]]
  Category Load (1.2ms)  SELECT  "categories".* FROM "categories" WHERE "categories"."id" = $1 LIMIT $2  [["id", 1693], ["LIMIT", 1]]
  SQL (1.2ms)  UPDATE "categories" SET "engagement_score" = COALESCE("engagement_score", 0) - 1, updated_at = '2017-12-07 09:16:36' WHERE "categories"."id" = $1  [["id", 1693]]
  Post Load (1.4ms)  SELECT  "posts".* FROM "posts" WHERE (posts.id = 64518) ORDER BY "posts"."id" ASC LIMIT $1  [["LIMIT", 1]]
  SQL (1.3ms)  UPDATE "posts" SET "engagement_score" = COALESCE("engagement_score", 0) - 1, updated_at = '2017-12-07 09:16:36' WHERE "posts"."id" = $1  [["id", 64518]]
  Post Load (1.3ms)  SELECT  "posts".* FROM "posts" WHERE (posts.id = 64518) ORDER BY "posts"."id" ASC LIMIT $1  [["LIMIT", 1]]
  SQL (1.5ms)  UPDATE "posts" SET "upvote_count" = COALESCE("upvote_count", 0) - 1 WHERE "posts"."id" = $1  [["id", 64518]]
   (2.1ms)  COMMIT
```
so I split `was` option into `relation_was_changed` and `delta_was_changed`

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -27,30 +27,28 @@ module CounterCulture
     #   :counter_cache_name => the column name of the counter cache
     #   :counter_column => overrides :counter_cache_name
     #   :delta_column => override the default count delta (1) with the value of this column in the counted record
-    #   :was => whether to get the current value or the old value of the
+    #   :relation_was_changed => whether to get the current value or the old value of the
     #      first part of the relation
     #   :execute_after_commit => execute the column update outside of the transaction to avoid deadlocks
     #   :with_papertrail => update the column via Papertrail touch_with_version method
-    def change_counter_cache(obj, options)
-      change_counter_column = options.fetch(:counter_column) { counter_cache_name_for(obj) }
-
+    def change_counter_cache(obj, increment: true, relation_was_changed: false, delta_was_changed: false, counter_column: counter_cache_name_for(obj), **options)
       # default to the current foreign key value
-      id_to_change = foreign_key_value(obj, relation, options[:was])
+      id_to_change = foreign_key_value(obj, relation, relation_was_changed: relation_was_changed)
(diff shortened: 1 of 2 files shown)


magnusvk:
Thanks, and sorry for the delay here. However, I don't think this is the right approach to avoid these queries—if we wanted to avoid these, shouldn't we try to use the Rails cache for relations that are already loaded? counter_culture always issues a `SELECT` from [here](https://github.com/magnusvk/counter_culture/blob/master/lib/counter_culture/counter.rb#L127) rather than using the relation which I think is causing this.

Ultimately, I'm not sure this is worth doing, but if we did, I'd want to try using the relation method here rather than using `where()` which will always issue a `SELECT` (though then perhaps we'd have to worry about stale associations).

Going to close this for now as I wouldn't want to merge it as is.

Issue #210, opened 2018-01-25
Set updated_at when executing counter_culture_fix_counts


AlexVPopov:
```ruby
class Foo < AR::Base
  has_many :bars
end

class Bar < AR::Base
  belongs_to :foo
  counter_culture :foo, touch: true
end
```

and

```ruby
Bar.counter_culture_fix_counts
```

I think in this case also the `updated_at` field of the `foos` needs to be updated. 

Does this make sense to you?

AlexVPopov:
Currently, as I workaround, in a migration, I'd put `Foo.find_each(&:touch)`.

magnusvk:
Yeah, that makes sense. Feel free to submit a PR. 😄

AlexVPopov:
@magnusvk I'm willing to submit a PR. What I've done so far:
* forked the project;
* cloned locally the fork;
* ran `rspec spec`

I get the following failure:

```bash
Failures:

  1) CounterCulture should fix a STI counter cache correctly
     Failure/Error: expect(company.twitter_reviews_count).to eq(1)

       expected: 1
            got: 0

       (compared using ==)
     # ./spec/counter_culture_spec.rb:1018:in `block (2 levels) in <top (required)>'

Finished in 2.31 seconds (files took 1.55 seconds to load)
35 examples, 1 failure

Failed examples:

rspec ./spec/counter_culture_spec.rb:1003 # CounterCulture should fix a STI counter cache correctly
```

After some investigation I found out that the failure is cause by [the `after_create` callback in `Review`](https://github.com/magnusvk/counter_culture/blob/master/spec/models/review.rb#L20). When I comment out this callback, the test passes. Also, if I put the callback after:

```ruby

# TwitterReview
counter_culture [:user, :manages_company]
```

that is, I move it from `Review` to `TwitterReview` (for the sake of executing it after counter_culter has finished updating the counters) the test also passes. I am currently investigating this, but it appears that after `after_create :update_some_text` is ran, `@_counter_culture_active` is not properly set to `false` in `_wrap_in_counter_culture_active(&block)`, so that when it gets to decide whether to update the counter for `twitter_reviews_count` in Company, `@_counter_culture_active` is still `true` and thus does not execute the block.

Before I continue my investigation - can you tell me if the test fails on your side as well?

I'm running ruby 2.4.1 and rails 5.1.5.

https://github.com/magnusvk/counter_culture/blob/master/lib/counter_culture/extensions.rb#L78

[...]

PR #212, opened 2018-02-21
Allow updating of timestamps when fixing counts

AlexVPopov:
As discussed in #210, I'm opening a PR to allow updating of timestamps for `counter_culture_fix_counts`. I appreciate your feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,18 @@ Product.counter_culture_fix_counts batch_size: 100
 
 ```counter_culture_fix_counts``` is optimized to minimize the number of queries and runs very quickly.
 
+Similarly to `counter_culture`, it is possible to update the records' timestamps, when fixing counts. If you would like to update the default timestamp field, pass `touch: true` option:
+
+```ruby
+Product.counter_culture_fix_counts touch: true
+```
+
+If you have specified a custom timestamps column, pass its name as the value for the `touch` option:
+
+```ruby
+Product.counter_culture_fix_counts touch: category_count_changed
+```
+
 #### Handling dynamic column names
 
 Manually populating counter caches with dynamic column names requires additional configuration:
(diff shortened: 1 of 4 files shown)


magnusvk:
@AlexVPopov this looks great, thank you! Sorry for the slow turnaround on review.

Just two things:
1. Can you rebase / merge in latest master to get rid of that test failure?
2. Can you add a test to make sure the timestamps don't change when `touch` isn't passed?

Thanks again!

AlexVPopov:
@magnusvk I rebased on master. Sure, it is no problem for me to add the test you mentioned, but I think that the the first test I added already covers this:

```ruby
...
Review.counter_culture_fix_counts(skip_unsupported: true, only: :user)
Review.counter_culture_fix_counts(skip_unsupported: true, only: :product, touch: true)
...
expect(user.updated_at).to eq old_user_updated_at
expect(product.updated_at).to be > old_product_updated_at
...
```

If you still think it makes sense, I'll add a separate test for this.

AlexVPopov:
Hey @magnusvk No problem, I agree it's nicer to have two tests about that. PR updated.

By the way, would you have accepted the following:

```ruby
Timecop.travel(1.second.from_now) do
  Review.counter_culture_fix_counts(skip_unsupported: true, only: :product, touch: true)

  expect { product.reload }.to change { product.updated_at }
end
```
?

On the one hand, it does not assert that the new `updated_at` is after the old one, but on the other hand, it makes assigning `old_product_updated_at` redundant and reads better IMO.

magnusvk:
@AlexVPopov thank you, merged. I think I prefer the current test where it explicitly asserts that it's later, but yeah ^^ that also would have worked.

magnusvk:
@AlexVPopov released as 1.10.0. Thanks again!

AlexVPopov:
@magnusvk Glad I could contribute.

PR #214, opened 2018-03-04
add better support of paranoia gem

epoberezhny:


magnusvk:
Thank you for this @epoberezhny — can you provide some more background on this, though? I don't quite understand the issue this is addressing yet.

epoberezhny:
Thank you for response! Earlier if we caused the destroying on the same record more than once with paranoia gem, the counter changed more than once too.

magnusvk:
@epoberezhny thank you!

magnusvk:
@epoberezhny just released this as 1.9.2

epoberezhny:
Cool, thanks @magnusvk! I also noticed that it can be used without the `model` argument in lambdas if you want a more compact code.

PR #218, opened 2018-05-02
Use PaperTrail's new save_with_version when available

### LINKED ISSUE #216: with_papertrail broken when using Papertrail 9x
The behavior of this option and its intent is now broken in Papertrail 9 due to this issue:
https://github.com/airblade/paper_trail/issues/1076

When using the `with_papertrail` option and a change is made which affects a counter_culture model, the model's counted row is *not* updated in the papertrail versions table.

magnusvk:
See https://github.com/airblade/paper_trail/issues/1076

Fixes #216

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -71,8 +71,9 @@ module CounterCulture
           if @with_papertrail
             instance = klass.where(primary_key => id_to_change).first
             if instance
-              if PaperTrail::VERSION::MAJOR >= 9
-                instance.touch
+              if instance.paper_trail.respond_to?(:save_with_version)
+                # touch_with_version is deprecated starting in PaperTrail 9.0.0
+                instance.paper_trail.save_with_version(validate: false)
               else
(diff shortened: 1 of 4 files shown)


key88sf:
👍

lib/counter_culture/counter.rb
```diff
@@ -68,18 +68,19 @@ def change_counter_cache(obj, options)
           klass = relation_klass(relation, source: obj, was: options[:was])
           primary_key = relation_primary_key(relation, source: obj, was: options[:was])
 
+          klass.where(primary_key => id_to_change).update_all updates.join(', ')
```
key88sf: @magnusvk I don't think we want to move this before the papertrail call, because that defeats the purpose of creating the version row. It's precisely because the `update_all` does not trigger any version rows to be created which is why we're forcing one before the update happens (in order to record the state of the object before it gets updated).

In our local specs that test this behavior, we compare counted_class.versions.last.object["counted_column_name"] to what it was *before* the update_all.

PR #220, opened 2018-05-31
Support Discard for soft-deletes

reidab:
This adds support for properly counting things when using the [Discard](https://github.com/jhawthorn/discard) gem for soft deletes.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,9 +26,18 @@ group :development, :test do
   gem "awesome_print"
   gem "timecop"
 
-  # to test the integration
-  gem "discard"
-  gem "paranoia"
+  # To test integrations
+
+  # Both the paranoia and discard integrations require Rails > 4.2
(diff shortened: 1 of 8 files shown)


jhawthorn:
I'm not sure this makes sense unless a `default_scope` is used (which isn't default or recommended).

reidab:
Can you elaborate? I added this patch specifically because I didn't want to use `default_scope` with Discard.

As near as I can tell, this works fine without the default scope by:

- decrementing the count `before_discard`
- incrementing it `before_undiscard`
- filtering on records where the `discard_column` is not null when reconciling counts

magnusvk:
@reidab this makes sense to me; a soft-deleted model should still decrement the count. Released as 1.12.0. Thanks for your contribution!

Also @jhawthorn 👋 — small world!

jhawthorn:
:wave: :heart: @magnusvk Hope you're doing well!

PR #227, opened 2018-08-19
Re-opening PR 224 here so that I can push to it

### LINKED ISSUE #224: Timestamps not set when creating papertrail version row from counter update
@magnusvk Another bug from using the new `save_with_version` function from Papertrail. The instance's `updated_at` is not being updated now with this new method. 

Papertrail uses the `updated_at` value of the instance to set the `created_at` timestamp on their new version row (presumably so it exactly matches the time the model itself was updated).

So now what we have happening is that version rows are created for the counter updates, but their `created_at` timestamps are not indicative of when the row was actually created. They are just set to whatever value happened to be in the `updated_at` column for the model itself.

I believe the fix for this is to simply set the `updated_at` of the instance inside of the `if instance.paper_trail.respond_to?(:save_with_version)` block, similar to what you do for "touch":

```
            if instance.paper_trail.respond_to?(:save_with_version)
              # touch_with_version is deprecated starting in PaperTrail 9.0.0
              current_time = instance.send(:current_time_from_proper_timezone)
              instance.send(:timestamp_attributes_for_update_in_model).each do |column|
                instance.send("#{column}=", current_time)
              end

              instance.paper_trail.save_with_version(validate: false)
            else
              instance.paper_trail.touch_with_version
            end
```

magnusvk:
Fixes #224, fixes #225


magnusvk:
@key88sf alright, this would be ready to merge. But I'm thinking about this now: Is this expected behavior? Basically, you are now implying `touch: true` when you set `papertrail: true`, right? That might be fine, and I think it's consistent with previous behavior, but maybe it should be something that's at least documented? Anyhow, I think this is probably fine, but just want to double-check with you on that before I merge and release.

key88sf:
@magnusvk Yes that sounds right. It's kind of an unfortunate artifact of how paper_trail uses the timestamp columns instead of setting the timestamp of the version row based on the current time.

PR #234, opened 2018-10-18
Test that it works with a model that has a module

### LINKED ISSUE #232: Gem does not properly deduce namespaced classes database table name
Overview:

I have two classes -- call them `FooBar::Show` and `FooBar::Act`.  Each show has many acts.  Show has a column `foo_bar_act_count` (defined with `null: false, default: 0`).  Add 'counter_culture :show` to the act class to tell Counter Culture to maintain my count column.

Expected Behaviour:

Running `FooBar::Act.counter_culture_fix_counts` should update the count values in the various Show records, with counter culture recognizing that the table name should be `foo_bar_act_count`.

Actual Behaviour:

Running `FooBar::Act_counter_culture_fix_counts` generates an error about `foo_bar_shows.foo_bar` does not exist.  Investigating the actual query, it is attempting to reference `"foo_bar_shows".foo_bar/acts_count`, a clearly invalid table name.

Workaround:

I can work around this easily by defining the counter column (`counter_culture :show, column_name: 'foo_bar_acts_count'`), but it seems odd that I should have to.  If the gem had generated a valid column name I'd consider shrugging it off, but the fact that the gem produces a column name that is, on it's face, invalid, suggests there is a genuine bug here.

magnusvk:
Fixes #232

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -13,7 +13,7 @@ module CounterCulture
         fail("execute_after_commit was removed; updates now run within the transaction")
       end
 
-      @counter_cache_name = options.fetch(:column_name, "#{model.name.tableize}_count")
+      @counter_cache_name = options.fetch(:column_name, "#{model.name.demodulize.tableize}_count")
       @column_names = options[:column_names]
       @delta_column = options[:delta_column]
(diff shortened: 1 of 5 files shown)


magnusvk:
Released as 2.1.0.

PR #237, opened 2018-11-06
Don't update counts when hard-destroying discarded records

tr4b4nt:
Hi,

this fixes bug when calling destroy on already discarded record (when using Discard for soft-deletes) - counts are currently updated twice (once when discard is called and once when destroy is called).


magnusvk:
@tr4b4nt thank you! Released as 2.1.1.

PR #244, opened 2019-01-17
Don't update running total on soft-deleted records

tr4b4nt:
Hi,

this fixes bug when using running total with Discard/Paranoia. 
Running total of soft-deleted records is currently updated when value of delta_column is changed.

### DIFF
--- a/lib/counter_culture/extensions.rb
+++ b/lib/counter_culture/extensions.rb
@@ -37,7 +37,9 @@ module CounterCulture
           end
 
           after_update :_update_counts_after_update, if: -> (model) do
-            if defined?(Discard::Model) && model.class.include?(Discard::Model)
+            if model.respond_to?(:paranoia_destroyed?)
+              !model.paranoia_destroyed?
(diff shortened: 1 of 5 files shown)


lib/counter_culture/extensions.rb
```diff
@@ -36,7 +36,15 @@ def counter_culture(relation, options = {})
...
-          after_update :_update_counts_after_update
+          after_update :_update_counts_after_update, if: -> (model) do
+            if model.respond_to?(:paranoia_destroyed?)
+              !model.paranoia_destroyed?
+            elsif defined?(Discard::Model) && model.class.include?(Discard::Model)
+              !model.discarded?
+            else
+              true
+            end
+          end
```
magnusvk: Can you extract this into a method please? Call it `destroyed_for_counter_culture?` perhaps (and use it with an `unless`)—that way it would be easy to override.

tr4b4nt:
Merging master solved test failures, thanks.

I added `destroyed_for_counter_culture?`.

PR #247, opened 2019-03-08
Fixed "size logic" to use the "cache counter column" instead of hitting the database (reopens #242)

dgilperez:
This PR reopens https://github.com/magnusvk/counter_culture/pull/242, that was closed, discussion mentioned reasons for reopening but is not done due to some Github permissions issue. 

This is rebased against `master`. 

I also think this is a major think to fix in this library, in order to keep with expected behaviour in a Rails app. `size` vs `count` is a well known feature (deciding on SQL or use a cache in the first, forcing a `SQL` count every time with the latter) and without that working, the power of counter caches is half-lost and results are unexpected. 

I had a deep look at `ActiveRecord::Associations::HasManyAssociation` and `ActiveRecord::Associations::CollectionAssociation`, the classes responsible for `size` logic. `size` is implemented like this

```ruby
# ActiveRecord::Associations::CollectionAssociation

def size
  if !find_target? || loaded?
    target.size
  elsif !association_scope.group_values.empty?
    load_target.size
  elsif !association_scope.distinct_value && target.is_a?(Array)
    unsaved_records = target.select(&:new_record?)
    unsaved_records.size + count_records
  else
    count_records
  end
end
```

In the first 3 branches are out of our scope (ie they do not use counter cache feature). In the last, `count_records` is implemented in `ActiveRecord::Associations::HasManyAssociation` as: 

```ruby
# ActiveRecord::Associations::HasManyAssociation

private 

def count_records
  count = if reflection.has_cached_counter?
    owner._read_attribute(reflection.counter_cache_column).to_i
  else
    scope.count(:all)
  end

  # If there's nothing in the database and @target has no new records
  # we are certain the current target is an empty array. This is a
  # documented side-effect of the method that may avoid an extra SELECT.
  (@target ||= []) && loaded! if count == 0

  [association_scope.limit_value, count].compact.min
end
```

So there are mainly two options: 

* To wrap `count_records` as this PR does.
* To wrap `ActiveRecord::Reflection::AbstractReflection#has_cached_counter?` and potentially `ActiveRecord::Reflection::AbstractReflection#counter_cache_column`. To be explored. Both are public methods. 

I can see reasons in favour of any of those. The implemented one seems to work fine (being tested in a production app as we speak). 

How do you want to proceed?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,7 +52,7 @@ group :development do
 end
 
 group :test do
-  gem "sqlite3"
+  gem "sqlite3", "~> 1.3.6"
   gem "rspec-extra-formatters"
   gem "database_cleaner", ">= 1.1.1"
 end
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -22,7 +22,7 @@ ActiveSupport.on_load(:active_record) do
   include CounterCulture::Extensions
   ActiveRecord::Associations::HasManyAssociation.send :prepend, CounterCulture::ActiveRecord::Associations::HasManyAssociation
 
-  if Rails.version >= '5.0'
+  if Rails.version < '4.2.0'
     ActiveRecord::Reflection::AssociationReflection.send :include, CounterCulture::ActiveRecord::Reflection::HasManyReflection
   else
     ActiveRecord::Reflection::HasManyReflection.send :include, CounterCulture::ActiveRecord::Reflection::HasManyReflection
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -21,5 +21,10 @@ end
 ActiveSupport.on_load(:active_record) do
   include CounterCulture::Extensions
   ActiveRecord::Associations::HasManyAssociation.send :prepend, CounterCulture::ActiveRecord::Associations::HasManyAssociation
-  ActiveRecord::Reflection::HasManyReflection.send :include, CounterCulture::ActiveRecord::Reflection::HasManyReflection
+
+  if Rails.version >= '5.0'
+    ActiveRecord::Reflection::AssociationReflection.send :include, CounterCulture::ActiveRecord::Reflection::HasManyReflection
+  else
+    ActiveRecord::Reflection::HasManyReflection.send :include, CounterCulture::ActiveRecord::Reflection::HasManyReflection
+  end
 end
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -5,6 +5,7 @@ require 'active_support/lazy_load_hooks'
 require 'counter_culture/extensions'
 require 'counter_culture/counter'
 require 'counter_culture/reconciler'
+require 'counter_culture/active_record/associations/has_many_association'
 
 module CounterCulture
   mattr_accessor :batch_size
@@ -19,4 +20,5 @@ end
 # extend ActiveRecord with our own code here
 ActiveSupport.on_load(:active_record) do
   include CounterCulture::Extensions
+  ActiveRecord::Associations::HasManyAssociation.send :prepend, CounterCulture::ActiveRecord::Associations::HasManyAssociation
 end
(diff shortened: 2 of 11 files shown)


magnusvk:
Thank you for this @dgilperez. This already looks a lot cleaner. That being said, my inclination here would be to wrap `has_cached_counter?` and `counter_cache_column` instead—if I read the `count_records` method correctly that should then "just work" shouldn't it? Is there a downside to that I can't see? It seems a lot safer to override / wrap public methods than private and it also avoids a bit of copy&paste of the stuff in `count_records`.

Also, if this doesn't work in older Rails versions, don't worry about it—I should drop support for those anyhow, and even if I don't we can skip those specs in older Rails versions.

dgilperez:
Hi @magnusvk, thanks for reviewing! 

I see what you say. I've being tinkering with a possible refactor in this direction, but I found it's a bit more difficult than expected: since `has_cached_counter?` and `counter_cache_name` are defined in `ActiveRecord::Reflection::AbstractReflection`, a class that's [inherited by plenty](https://www.rubydoc.info/github/rails/rails/ActiveRecord/Reflection/AbstractReflection), I've had a hard time fixing all potential cases. 

The results of this refactor are in a new branch / PR #248. Specs are green, but my app will refuse to load with that code. The code in this PR is working in production though. I kind of prefer the approach on #248, since it's wrapping only public methods, but on the other hand I find this implementation more focused. Both approaches tinker with quite a lot of internals. The private method wrapped here has being stable for several versions now, I don't expect it to change a lot either. 

I don't expect to have a lot of time in the coming weeks to look further into this though :( so we're going to use this very code (not the one in #248) in our app until this issue is finally sorted out. 

Let me know your thoughts.

Best!

magnusvk:
You're saying your app flat-out refuses to load with #247? What's the error you get there? 

Seeing both approaches side-by-side I definitely prefer the version that wraps only public methods, so it's a bummer to hear that's not working in your app.

dgilperez:
Yes, it's a bummer :( Application won't load in deed. The error of trying to run `rspec spec/models/user_spec.rb` 

> <rvm_path>/gems/activerecord-5.2.2.1/lib/active_record/dynamic_matchers.rb:22:in `method_missing': undefined method `devise' for User (call 'User.connection' to establish a connection):Class (NoMethodError)
>	from <code_path>/app/models/user.rb:5:in `<class:User>'

At that line you find the first method call, `devise ...`

I assumed I went too deep into AR internals, and stop going down the path. I also have this 247 PR working in production at this moment, so this is now not a top priority for us.

I still think this is a major issue that counter_culture should address and fix. Let me know if I can be of any extra help.

magnusvk:
Hm. This looks like something we're doing in counter_culture is triggering ActiveRecord to initialize and connect to the DB before things are ready. There is probably some logic that needs to move from a place that gets called right when the code is first read to a later point, post-init. Hard to tell what that is though without more info. Perhaps with the full stack-trace.

But in any case it sounds like neither of us really has a ton of time to get this to be production-ready for inclusion in the gem here. That's ok for me, but just want to be upfront that I don't have a ton of time to make this work either.

Thanks for contributing this though!

dgilperez:
I agree and thanks for the clear answer. It's correct that we don't have a ton of time for this now, and probably will not in the next weeks :(. At the same time we are using that other PR in some projects already, so we're considering our fork production-ready. 

I perfectly understand that it's not the change you want to include here, so I guess we'll keep rebasing our fork until we free some time on our end or this gets fixed here. Please close the PR (or both) if you feel this is going to get stale. 

At least, I think current behaviour of `size` should be noted in the README or somewhere else as a known gotcha, to prevent unexpected results for others.

magnusvk:
I'm happy to keep the PRs open for now. And I've updated the Readme here: https://github.com/magnusvk/counter_culture/commit/80f62ada3ce79efce47f3ec5357f51b15350fdb1

magnusvk:
Closing this as it's now old.

PR #248, opened 2019-03-17
Refactors #247 - Fixed size logic

dgilperez:
This is a refactor from #247, work in progress. 

The reason for opening a new PR is that #247 implementation is working properly in a production app, while this code is not (although specs are green) and because I think it makes sense to consider both approaches.

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -1,6 +1,6 @@
 module CounterCulture
   class Counter
-    CONFIG_OPTIONS = [ :column_names, :counter_cache_name, :delta_column, :foreign_key_values, :touch, :delta_magnitude]
+    CONFIG_OPTIONS = [ :column_names, :counter_cache_name, :delta_column, :foreign_key_values, :touch, :delta_magnitude, :condition ]
     ACTIVE_RECORD_VERSION = Gem.loaded_specs["activerecord"].version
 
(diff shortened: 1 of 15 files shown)


lib/counter_culture/counter.rb
```diff
@@ -1,6 +1,6 @@
 module CounterCulture
   class Counter
-    CONFIG_OPTIONS = [ :column_names, :counter_cache_name, :delta_column, :foreign_key_values, :touch, :delta_magnitude]
+    CONFIG_OPTIONS = [ :column_names, :counter_cache_name, :delta_column, :foreign_key_values, :touch, :delta_magnitude, :condition ]
```
magnusvk: Separate question, but is this change unrelated? I'm a little confused as to the purpose of this.

dgilperez: You're right, this looks like another unrelated fix, related with the ability of passing `if:` to `:counter_culture`. @WaKeMaTTa can you add a bit more here, please?

phucbnh-1838: in my opinion you can write this as follow: `CONFIG_OPTIONS = %i[ column_names counter_cache_name delta_column foreign_key_values touch delta_magnitude condition ]`

magnusvk:
Closing this as it's now old.

PR #250, opened 2019-04-03
Implement where option to Reconcile

ChrisBr:
to make it possible to limit the number of records which get updated.

For instance, we have quite a big number of records and running periodically on all records might not work so it would make sense to run periodically only on a subset of the records and schedule it to different times.

Does this make sense?


magnusvk:
@ChrisBr this looks great, thank you. Can you merge in the latest master to get the specs green before I merge?

ChrisBr:
@magnusvk rebased 👍

magnusvk:
@ChrisBr thanks for this, too! Will go out with 2.2.0 as well.

PR #251, opened 2019-04-04
Add verbose mode to Reconciler

ChrisBr:
to print some logs to STDOUT.

### DIFF
--- a/lib/counter_culture/reconciler.rb
+++ b/lib/counter_culture/reconciler.rb
@@ -67,6 +67,7 @@ module CounterCulture
       end
 
       def perform
+        log "Performing reconciling of #{counter.model}##{counter.relation.to_sentence}.\n"
         # if we're provided a custom set of column names with conditions, use them; just use the
         # column name otherwise
         # which class does this relation ultimately point to? that's where we have to start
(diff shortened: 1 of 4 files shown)


lib/counter_culture/reconciler.rb
```diff
@@ -101,12 +102,23 @@ def perform
...
         end
       end
 
       private
 
+      def log
+        @log ||= if options[:verbose]
+                   Logger.new(STDOUT)
+                 else
+                   Logger.new('/dev/null')
```
magnusvk: I don't love this; some thoughts / questions:
1. Why not use `Rails.logger` here, at least if it's available?
2. I wouldn't use a logger to `/dev/null`, I think I'd prefer us to encapsulate this with a `log(message)` method that logs only if verbose is true.

What do you think?

ChrisBr:
@magnusvk thanks for your valuable feedback. I changed the code and also added a spec for it. Please have a look at it again and let me know what you think 👀

magnusvk:
@ChrisBr thanks! Going to release this as 2.2.0 when the specs go green.

PR #254, opened 2019-04-09
Use a MAX() aggregation when fetching current count during reconciliation

### LINKED ISSUE #253: PostgreSQL error when primary_key column is not indexed as primary_key
On PostgreSQL, if the table where the counts are stored doesn't have a primary_key specified (in our case, because it's actually a view), reconciliation fails with an error:

```
PG::GroupingError:
   ERROR:  column "foo.users_count" must appear in the GROUP BY clause or be used in an aggregate function
   LINE 1: ...foo".id, COUNT("users".id)*1 AS count, "foo..
```

This is because, according to the [Postgres docs](https://www.postgresql.org/docs/current/static/sql-select.html#SQL-GROUPBY):

> When GROUP BY is present, or any aggregate functions are present, it is not valid for the SELECT list expressions to refer to ungrouped columns except within aggregate functions or when the ungrouped column is functionally dependent on the grouped columns, since there would otherwise be more than one possible value to return for an ungrouped column. A functional dependency exists if the grouped columns (or a subset thereof) are the primary key of the table containing the ungrouped column.

When there's no primary key index present, even though Rails knows what the primary key is, no functional dependency gets detected at the DB level. 

I'm putting together a PR to address this by wrapping the fetch for the count column in a MAX aggregate function, which should effectively always be a no-op.

reidab:
Fixes #253

On PostgreSQL, if the table where the counts are stored doesn't have a primary_key specified (in our case, because it's actually a view), reconciliation fails with an error:

```
PG::GroupingError:
   ERROR:  column "foo.users_count" must appear in the GROUP BY clause or be used in an aggregate function
   LINE 1: ...foo".id, COUNT("users".id)*1 AS count, "foo..
```

This is because, according to the [Postgres docs](https://www.postgresql.org/docs/current/static/sql-select.html#SQL-GROUPBY):

> When GROUP BY is present, or any aggregate functions are present, it is not valid for the SELECT list expressions to refer to ungrouped columns except within aggregate functions or when the ungrouped column is functionally dependent on the grouped columns, since there would otherwise be more than one possible value to return for an ungrouped column. A functional dependency exists if the grouped columns (or a subset thereof) are the primary key of the table containing the ungrouped column.

When there's no primary key index present, even though Rails knows what the primary key is, no functional dependency gets detected at the DB level. 

This `MAX` agg should effectively be a no-op in every case, but it prevents PG from blowing up. An alternate fix to this same problem would be adding `"#{relation_class_table_name}.#{column_name}"` to the GROUP BY fields.

**Note:** the specs added in this commit will pass on SQLite even without these fixes — they work as expected when running the specs on a PostgreSQL DB. Running the tests on PG also reveals that the existing string ID specs (which also don't have a primary_key index) fail in the same way).

### DIFF
--- a/lib/counter_culture/reconciler.rb
+++ b/lib/counter_culture/reconciler.rb
@@ -86,7 +86,12 @@ module CounterCulture
           relation_class_table_name = quote_table_name(relation_class.table_name)
 
           # select join column and count (from above) as well as cache column ('column_name') for later comparison
-          counts_query = scope.select("#{relation_class_table_name}.#{relation_class.primary_key}, #{relation_class_table_name}.#{relation_reflect(relation).association_primary_key(relation_class)}, #{count_select} AS count, #{relation_class_table_name}.#{column_name}")
+          counts_query = scope.select(
+            "#{relation_class_table_name}.#{relation_class.primary_key}, " \
+            "#{relation_class_table_name}.#{relation_reflect(relation).association_primary_key(relation_class)}, " \
+            "#{count_select} AS count, " \
+            "MAX(#{relation_class_table_name}.#{column_name}) AS #{column_name}"
+          )
 
           # we need to join together tables until we get back to the table this class itself lives in
           join_clauses(where).each do |join|
--- a/spec/counter_culture_spec.rb
+++ b/spec/counter_culture_spec.rb
@@ -9,6 +9,7 @@ require 'models/twitter_review'
 require 'models/user'
 require 'models/category'
 require 'models/has_string_id'
+require 'models/has_non_pk_id'
 require 'models/simple_main'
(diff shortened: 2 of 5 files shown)


reidab:
Any chance of getting this merged soon? Would it be helpful for me to set up PostgreSQL testing on Travis to demonstrate the issue?

magnusvk:
Yes, this makes sense, thank you for this. Sorry, I didn't get a notification from Github when you opened this for some reason.

Released this as 2.2.2 right now.

magnusvk:
Thinking about this some more: Do you think you could get this test running in PostgreSQL also? I think that would be really good to make sure I don't break this for you again in the future.

PR #260, opened 2019-06-04
Fix counter_culture option name of README.md

umemotoryo:
I'm using it but option name is invalid.
So I fixed option name of Usage of README.md.


magnusvk:
Thank you!

PR #261, opened 2019-06-19
Use klass.connection instead to make sure we get the correct connection

magnusvk:
... in case different models have different database connections.


magnusvk:
cc @reidab

reidab:
LGTM 👍

magnusvk:
Awesome, I released your changes as 2.2.3 to get the MySQL fixes out there. Thanks again for doing this, that setup is definitely a huge step forward! Used it to quickly add Rails 6 testing, too.

PR #266, opened 2019-08-10
Duplicate decrement occurs under specific conditions (race condition bug)

soranoba:
Thank you for a good library.

I have been suffering from a bug that the counter breaks.
The cause has not been identified, but I confirmed that the counter would break under certain conditions.

----

If users try to delete the same record by making DELETE requests at the same time, the counter will be decremented at the some times.
Please see [this commit](https://github.com/soranoba/counter_culture/commit/fe7d0fe1ac81012ece3779a6e3dcdab43afffdd9) and [result of CI](https://travis-ci.org/soranoba/counter_culture/builds/570116324).

This is because `destroy` (or `destroy!`) does not throw an exception if the record has already been deleted.
Also, `after_destroy` is called as many times as `destroy` (or `destroy!`) is executed, but `after_commit on: :destroy` is called as many times as it actually deletes. (In other words, it is called only once)

This eliminates race condition issues.

Thanks!!

### DIFF
--- a/lib/counter_culture/extensions.rb
+++ b/lib/counter_culture/extensions.rb
@@ -20,27 +20,27 @@ module CounterCulture
           # initialize callbacks only once
           after_create :_update_counts_after_create
 
-          after_commit :_update_counts_after_destroy, on: :destroy, unless: :destroyed_for_counter_culture?
-
-          if respond_to?(:before_real_destroy) &&
-              instance_methods.include?(:paranoia_destroyed?)
-            before_real_destroy :_update_counts_after_destroy,
-              if: -> (model) { !model.paranoia_destroyed? }
-          end
-
-          after_update :_update_counts_after_update, unless: :destroyed_for_counter_culture?
-
-          if respond_to?(:before_restore)
+          if instance_methods.include?(:paranoia_destroyed?)
             before_restore :_update_counts_after_create,
-              if: -> (model) { model.deleted? }
-          end
-
-          if defined?(Discard::Model) && include?(Discard::Model)
+              if: ->(model) { model.deleted? }
+            after_update :_update_counts_after_update,
+              if: ->(model) { !model.paranoia_destroyed? }
+            before_real_destroy :_update_counts_after_destroy,
+              if: ->(model) { !model.paranoia_destroyed? }
+            before_destroy :_update_counts_after_destroy,
+              if: ->(model) { !model.paranoia_destroyed? }
+          elsif defined?(Discard::Model) && include?(Discard::Model)
             before_discard :_update_counts_after_destroy,
               if: ->(model) { !model.discarded? }
-
             before_undiscard :_update_counts_after_create,
               if: ->(model) { model.discarded? }
+            after_update :_update_counts_after_update,
+              if: ->(model) { !model.discarded? }
+            after_commit :_update_counts_after_destroy, on: :destroy,
+              if: ->(model) { !model.discarded? }
+          else
+            after_update :_update_counts_after_update
+            after_commit :_update_counts_after_destroy, on: :destroy
           end
 
           # we keep a list of all counter caches we must maintain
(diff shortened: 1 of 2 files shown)


soranoba:
It seems that the behavior is strange using paranoia.
Can someone help me?

Or is this correction wrong?

spec/counter_culture_spec.rb
```diff
@@ -111,6 +111,39 @@
     expect(product.reviews_count).to eq(0)
   end
 
+  it "does not duplicate decrements counter cache on destroy" do
+    # NOTE: If Rails.version < 5.1.0, this test fails because rails bugs.
+    #       You should use Rails >= 5.1.0
+    next if Rails.version < "5.1.0"
```
soranoba: Rails 5.0 : https://travis-ci.org/soranoba/counter_culture/jobs/570116514
Rails 5.1 : https://travis-ci.org/soranoba/counter_culture/jobs/570116515

soranoba:
A fix has been completed to work correctly except for existing bugs.
refer: #267

lib/counter_culture/extensions.rb
```diff
@@ -20,27 +20,27 @@ def counter_culture(relation, options = {})
...
-          if respond_to?(:before_real_destroy) &&
-              instance_methods.include?(:paranoia_destroyed?)
-            before_real_destroy :_update_counts_after_destroy,
-              if: -> (model) { !model.paranoia_destroyed? }
-          end
-
-          after_update :_update_counts_after_update, unless: :destroyed_for_counter_culture?
-
-          if respond_to?(:before_restore)
+          if instance_methods.include?(:paranoia_destroyed?)
```
soranoba: Using paranoia, it no change in behavior.

lib/counter_culture/extensions.rb
```diff
@@ -20,27 +20,27 @@ def counter_culture(relation, options = {})
...
-
-          if defined?(Discard::Model) && include?(Discard::Model)
+              if: ->(model) { model.deleted? }
+            after_update :_update_counts_after_update,
+              if: ->(model) { !model.paranoia_destroyed? }
+            before_real_destroy :_update_counts_after_destroy,
+              if: ->(model) { !model.paranoia_destroyed? }
+            before_destroy :_update_counts_after_destroy,
+              if: ->(model) { !model.paranoia_destroyed? }
+          elsif defined?(Discard::Model) && include?(Discard::Model)
```
soranoba: Using Discard, it no change in behavior.

lib/counter_culture/extensions.rb
```diff
@@ -20,27 +20,27 @@ def counter_culture(relation, options = {})
...
             before_discard :_update_counts_after_destroy,
               if: ->(model) { !model.discarded? }
-
             before_undiscard :_update_counts_after_create,
               if: ->(model) { model.discarded? }
+            after_update :_update_counts_after_update,
+              if: ->(model) { !model.discarded? }
+            after_commit :_update_counts_after_destroy, on: :destroy,
+              if: ->(model) { !model.discarded? }
+          else
```
soranoba: The behavior has only changed when it does not use Discard and paranoia.

magnusvk:
@soranoba thanks for reporting this. My first instinct is to say this is more of a Rails bug than a counter_culture bug, but I'm also realistic enough to accept that we might want a workaround here.

That being said, I don't like moving the destroy callback to `after_commit`—we should keep the behavior consistent between create, update and destroy. Can you think of a way of fixing this that keeps the callback within the transaction?

soranoba:
@magnusvk Thank you for reply.

> Can you think of a way of fixing this that keeps the callback within the transaction?

The `after_destroy` that triggers within a transaction does not provide the desired behavior now.
So it seems difficult.
refer: https://travis-ci.org/soranoba/counter_culture/builds/575398039?utm_source=github_status&utm_medium=notification

> we should keep the behavior consistent between create, update and destroy.

Of course. However, I think that the PR does not break it.

1. When the transaction is failed and is executed rollback, `after_commit` never called. So, the transaction is success if `after_commit` called.
2. When `after_commit on: :destroy` called, no more changes will be made to the same record. Because, it no longer exists.
3. Since `counter_culture` uses `COALESCE` to increase or decrease the value, there is no need to worry about exclusive control of this operation unless there is a place to set the value directly.

What do you think?

magnusvk:
Your change would still move the `after_destroy` change of the counter cache outside of the transaction, which takes away transactional guarantees. I'm ok with adding a workaround to counter_culture—though you should report this as a bug to Rails, as well. But I won't add a change that moves just one of the operations outside of the transaction.

magnusvk:
Closing for lack of response.

PR #271, opened 2020-01-08
Feature/round off values

md-farhan-memon:
We encountered a case where we were totalling float column and the resultant sum was not rounded off/out of precision e.g. `0.1 + 0.2` results in `0.30000000000000004` instead of `0.3`. Since `update_all` being used for updating the sum, AR callbacks were not helping either to format the data before updating.
Thus, I added an option `:precision` where one can provide the integer so as to specify number of decimal places for rounding off the sum.
I have added test cases as well, are all green and updated readme.

I believe it would be a good feature to have because, even though it can be rounded off while reading (as done in the test cases earlier), one might need to store the rounded values itself.

We are already using it in our code base and thought would be nice if other people can benefit from it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,31 @@ The value in the counter cache will be the sum of the ```weight_ounces``` values
 
 The ```:delta_column``` option supports all numeric column types, not just ```:integer```. Specifically, ```:float``` is supported and tested.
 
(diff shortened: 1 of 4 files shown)


magnusvk:
Closing for lack of response.

Issue #245, opened 2019-01-18
counter_culture_fix_counts does not work properly with dynamic columns


MobileD2:
Hi,

I have dynamic column set for counter_culture but when I tried to run `Order.counter_culture_fix_counts` in rails console to update counters, it raises me an error:

`RuntimeError (Must provide :column_names option for relation [:order_group] when :column_name is a Proc; you may skip this relation with :skip_unsupported => true)
`

My setup: Rails 5.2, counter_culture 2.1.2
Order and OrderGroup have `close_base_amount` field

Order.rb

```
class Order < ApplicationRecord

  belongs_to :order_group, touch: true

  counter_culture :order_group, column_name: proc{ |model| model.opened? ? :open_orders_count : nil }

  scope :opened, ->{ where(arel_table[:close_base_amount].eq(0)) }

  def opened?
    close_base_amount.zero?
  end
}
end
```

OrderGroup.rb

```
class OrderGroup < ApplicationRecord

  has_many :orders

  scope :opened, ->{ where(arel_table[:close_base_amount].eq(0)) }

end
```

also I added a column to OrderGroup model by migration

```
class AddCountersToOrderGroups < ActiveRecord::Migration[5.2]
  def change
    add_column :order_groups, :open_orders_count, :integer, null: false, default: 0
  end
end
```

When I tried to add `column_names` to Order counter_culture configuration like this:

```
  counter_culture :order_group,
                  column_name: proc{ |model| model.opened? ? :open_orders_count : nil },
                  column_names: {
                      ["order_groups.close_base_amount = ?", 0] => :open_orders_count
                  }
```

it helps, but I think it does not make sense - I have to repeat all my conditions in `column_names` in SQL format. 

Could you help me to find a proper way to use counter_culture with dynamic column name?
Is there a way to use scope in `column_names` instead of SQL query?

Regards, Dmytro

magnusvk:
That's the price you pay for being able to fix counts in aggregate—it needs SQL-style conditions. This is all working as designed and documented.

MobileD2:
I found one solution to do that. It could be done by extracting where clause from scope via Arel:

```
counter_culture :order_group,
                  column_name: proc{ |model| model.opened? ? :open_orders_count : nil },
                  column_names: {
                      [Order.opened.arel.ast.cores.last.wheres.last.to_sql] => :open_orders_count
                  }
```

magnusvk:
[...]

PR #272, opened 2020-01-09
Extend column_names option to accepting scope name as symbol hash key

takuan-oishii:
Hello.
I extended column_names option whose hash key accepts scope names as symbol, inspired by #245.
When you apply counter to a model that has a complex condition, you can use this scope as column_names's condition instead of Arel or SQL-style condition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,20 +329,20 @@ class Product < ActiveRecord::Base
 end
 ```
 
-You can specify scope name instead of where condition string.
+You can specify scope instead of where condition string.
 
 ```ruby
 class Product < ActiveRecord::Base
   belongs_to :category
+  scope :awesomes, ->{ where "products.product_type = ?", 'awesome' }
+  scope :suckys, ->{ where "products.product_type = ?", 'sucky' }
+
   counter_culture :category,
       column_name: proc {|model| "#{model.product_type}_count" },
       column_names: {
-          awesomes: :awesome_count,
-          suckys: :sucky_count
+          Product.awesomes => :awesome_count,
+          Product.suckys => :sucky_count
       }
-  scope :awesomes, ->{ where "products.product_type = ?", 'awesome' }
-  scope :suckys, ->{ where "products.product_type = ?", 'sucky' }
-  # attribute product_type may be one of ['awesome', 'sucky']
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -329,6 +329,23 @@ class Product < ActiveRecord::Base
 end
 ```
 
+You can specify scope name instead of where condition string.
+
+```ruby
+class Product < ActiveRecord::Base
+  belongs_to :category
+  counter_culture :category,
+      column_name: proc {|model| "#{model.product_type}_count" },
+      column_names: {
+          awesomes: :awesome_count,
+          suckys: :sucky_count
+      }
+  scope :awesomes, ->{ where "products.product_type = ?", 'awesome' }
+  scope :suckys, ->{ where "products.product_type = ?", 'sucky' }
+  # attribute product_type may be one of ['awesome', 'sucky']
+end
+```
+
 If you would like to avoid this configuration and simply skip counter caches with
 dynamic column names, while still fixing those counters on the model that are not
 dynamic, you can pass `skip_unsupported`:
--- a/lib/counter_culture/reconciler.rb
+++ b/lib/counter_culture/reconciler.rb
@@ -252,8 +252,8 @@ module CounterCulture
           if index == reverse_relation.size - 1
             # conditions must be applied to the join on which we are counting
             if where
-              if where.is_a? Symbol
-                joins_sql += " AND #{target_table_alias}.#{model.primary_key} IN (#{model.send(where).pluck(model.primary_key).join(',')})"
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -329,6 +329,23 @@ class Product < ActiveRecord::Base
...
+You can specify scope name instead of where condition string.
+
+```ruby
+class Product < ActiveRecord::Base
+  belongs_to :category
+  counter_culture :category,
+      column_name: proc {|model| "#{model.product_type}_count" },
+      column_names: {
+          awesomes: :awesome_count,
+          suckys: :sucky_count
```
magnusvk: I like the approach of using scopes here, but could we use the actual scope here instead of symbols? That would then look like this:
```
  awesomes: Product.awesomes,
  suckys: Product.suckys
```

lib/counter_culture/reconciler.rb
```diff
@@ -252,7 +252,11 @@ def join_clauses(where)
           if index == reverse_relation.size - 1
             # conditions must be applied to the join on which we are counting
             if where
-              joins_sql += " AND (#{model.send(:sanitize_sql_for_conditions, where)})"
+              if where.is_a? Symbol
+                joins_sql += " AND #{target_table_alias}.#{model.primary_key} IN (#{model.send(where).pluck(model.primary_key).join(',')})"
```
magnusvk: I think that would simplify your work here, too, because you can get the SQL from the scope?

README.md
takuan-oishii: `column_names` option is defined as `{ 'where_cause' => 'counter_column_name'  }`. (I understood in this way. It is right?)
So I will extend the definition to `{ 'where_cause' or 'scope_name' => 'counter_column_name'  }`.


Sorry, I'm not sure what this mean.

```
  awesomes: Product.awesomes,
  suckys: Product.suckys
```

Do you wish extend the definition to `{ 'where_cause' => 'counter_column_name'  }` or `{ 'counter_column_name' => Model.calling_scope  }`?

If my guessing is right, I think the definition is confusing to user. 
I want to hear your opinion.


I'm not used to discussing English. I'm sorry if I'm rude.

magnusvk: No worries, not rude at all. And in my example I did it the wrong way around. What I meant was, could you use the actual scope instead of the name of the scope. So it would look like this:
```
column_names: {
           Product.awesomes => :awesome_count,
           Product.suckys => :sucky_count
}
```

I'm worried about the indirection of using a symbol. What if you put a symbol and then that doesn't exist as a scope? With your implementation that would just blow up on the `send` and be confusing. It seems to me that using the scope directly would be clearer and with less potential issues.

takuan-oishii: Wow! Its perfect idea!!

takuan-oishii: I implemented your idea. [Check this out](https://github.com/magnusvk/counter_culture/pull/272/commits/87c87be01efda567bb4a03ad2adf1aa62fc0bee0).

Chained scope becomes possible.

```
column_names: {
           Product.awesomes.suckys => :awesome_and_sucky_count
}
```

Its so good.

lib/counter_culture/reconciler.rb
takuan-oishii: I updated the code to this.

```
if where.respond_to?(:to_sql)
  joins_sql += " AND #{target_table_alias}.#{model.primary_key} IN (#{where.select(model.primary_key).to_sql})"
else
  joins_sql += " AND (#{model.send(:sanitize_sql_for_conditions, where)})"
end
```

I have no idea about simpler way.
Scope is able to contain JOIN cause, So I have to use IN cause and subquery.

spec/counter_culture_spec.rb
```diff
@@ -2285,6 +2287,17 @@ def mess_up_counts
     end
   end
 
+  it "can fix counts by scope" do
+    prefecture = Prefecture.new name: 'Tokyo'
+    prefecture.save!
+    City.create!(name: 'Sibuya', prefecture: prefecture, population: 221800)
+    City.create!(name: 'Oku Tama', prefecture: prefecture, population: 6045)
+
+    City.counter_culture_fix_counts
```
magnusvk: I think the value here will already be correct? Can you manually set the `big_cities_count` value to something incorrect?

lib/counter_culture/reconciler.rb
magnusvk: Yeah I think this is good.

spec/counter_culture_spec.rb
takuan-oishii: Exactly.
Its done. https://github.com/magnusvk/counter_culture/pull/272/commits/455e81f5d70d9b0f2adffb879b4687c28f1bd437

magnusvk:
Thank you for this. On the test, I meant setting the count to be incorrect after it had been counted up, so something like this: 831d2d0

But thank you, this is great! I'll release a new gem version once the tests are green.

takuan-oishii:
Aaaaaaaa I looked https://github.com/magnusvk/counter_culture/commit/831d2d098fe7da194269eee451588fd560d0ae9d and understood.
Thank you for fixing and merge!

PR #279, opened 2020-04-23
Add support for limits in batch processing

robertomiranda:
The options start and finish are especially useful if you want multiple workers dealing with the same processing queue. You can make worker 1 handle all the records between id 1 and 9999 and worker 2 handle from 10000 and beyond by setting the `:start` and `:finish` option on each worker.

Description is taken from [find_in_batches docs in rails](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_in_batches)

```ruby
Product.counter_culture_fix_counts start: 10_000
# will fix counts for all counter caches defined on Product from record 10000 and onwards.
Product.counter_culture_fix_counts finish: 10_000
# let's process until 10000 records.
Product.counter_culture_fix_counts start: 1000, finish: 2000
# In worker 1, lets process from 1000 to 2000
Product.counter_culture_fix_counts start: 2001, finish: 3000
# In worker 2, lets process from 2001 to 3000
```

Another use case of these options is to be able to restart the workers from specific id in case of failure in order to not revisit the records already processed

### DIFF
--- a/spec/counter_culture_spec.rb
+++ b/spec/counter_culture_spec.rb
@@ -2304,6 +2304,31 @@ RSpec.describe "CounterCulture" do
     expect(prefecture.reload.big_cities_count).to eq(1)
   end
 
+  it "support fix counts using batch limits start and finish" do
+    company = Company.create!
+    company.children << Company.create!
+    company.children_count = -1
+    company.save!
+
+    companies = 3.times.map do
+      company = Company.create!
+      company.children << Company.create!
+      company.children_count = -1
+      company.save!
+      company
+    end
+
+    start = companies.first.id
+    finish = companies.last.id
+
+    fixed = Company.counter_culture_fix_counts start: start, finish: finish
+    expect(fixed.length).to eq(3)
+
+    companies.each do |company|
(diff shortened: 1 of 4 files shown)


spec/counter_culture_spec.rb
```diff
@@ -2304,6 +2304,31 @@ def mess_up_counts
...
+
+    start = companies.first.id
+    finish = companies.last.id
+
+    fixed = Company.counter_culture_fix_counts start: start, finish: finish
+    expect(fixed.length).to eq(3)
+
+    companies.each do |company|
+      expect(company.reload.children_count).to eq(1)
+    end
```
magnusvk, line 2329: Should we also test that `company` _wasn't_ fixed here? To make sure that it's only touching the IDs that we're giving it? And then maybe pass `start` and `finish` that does include `company` and make sure it gets fixed on the second go around?

robertomiranda: I think I've addressed your feedback @magnusvk 👍

magnusvk:
Released as 2.4.0—thank you!

PR #281, opened 2020-05-09
Fix Rails version check in migration generator (Fixes #280)

### LINKED ISSUE #280: migration not Rails version aware
the generator creates a migration with the old-style Migration class `ActiveRecord::Migration`

```
Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:
```

I think there's a way to pipe the Rails version into the template as it is being generated 

I did it in another gem like this
```
class CreateUniversalTrackManagerTables < ActiveRecord::Migration<%= migration_version %>
  def self.up
    ActiveRecord::Base.transaction do

```

magnusvk:
To include the Rails version not just for Rails 5.x but also future
versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 ## 2.3.0 (January 28, 2020)
 
 Improvements:
+  - Allow specifying `start` and `finish` options to `counter_culture_fix_counts` (#279)
+
+## 2.3.0 (January 28, 2020)
+
+Improvements:
   - Allow using scopes in `column_names` (#272)
 
 ## 2.2.4 (August 21, 2019)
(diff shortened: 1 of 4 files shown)


jasonfb:
testing now

jasonfb:
works as expected. when I use this fix branch of this gem, the generator correctly generates a AR migration that has the version number

I tested just using the example in the README

```
MBA-gold:test-fix-rails-version-check-in-migration-test jason$ rails generate counter_culture Category products_count
Running via Spring preloader in process 92724
      create  db/migrate/20200517134855_add_products_count_to_categories.rb
```

Gemfile — test-fix-rails-version-check-in-migration-test 2020-05-17 09-51-21


20200517134855_add_products_count_to_categories rb — test-fix-rails-version-check-in-migration-test 2020-05-17 09-51-05

jasonfb:
that would be a 👍

magnusvk:
Thanks @jasonfb—fix released as 2.5.1.

PR #284, opened 2020-05-13
Remove unused variables

robertomiranda:
After merging https://github.com/magnusvk/counter_culture/pull/282 I think these two variables are no longer used 🧹


magnusvk:
Thank you, you are right of course. 🤦🏻‍♂️

Issue #278, opened 2020-04-08
counter_culture breaks code if any of the attributes is read only


Morozzzko:
I'm using [aasm](https://github.com/aasm/aasm) on my ActiveRecord model. I've used the `no_direct_assignment: true` option, so any attempts to call `my_model.attr = ...` will fail with an exception.

Once I introduce `counter_culture` to the class, the application breaks – the class will raise an exception if try to update it. 

This behavior should be reproducible even without AASM by overriding the writer, like this

```ruby
def attr=(*)
  raise "Welp"
end
```

**Here's why it happens**: 
1. `CounterCulture::Counter#previous_model` calls `#saved_changes` on my model, and it includes the read-only fields. It probably happens because the model has been created and updated in the same business process / transaction.  
2. It tries to write the attribute to recreate _some old_ state
3. It fails

Here's the problematic spot:

https://github.com/magnusvk/counter_culture/blob/e0e0302c56fb616d2d4cd8e5e93a035099629aeb/lib/counter_culture/counter.rb#L287-L297

Here's an old article on the same problem:

http://natural-hokke.hateblo.jp/entry/2017/12/29/221943

magnusvk:
Thanks for the detailed report—this makes sense. Are you able to open a pull request with a fix? If `aasm` uses the Rails-standard read-only attributes here, we should be able to just say `next if  obj.klass.readonly_attributes.include?(key)`, right? ([See here.](https://api.rubyonrails.org/classes/ActiveRecord/ReadonlyAttributes/ClassMethods.html#method-i-attr_readonly)) However, it will also need a test.

Morozzzko:
I'll need to check it out. I'm not sure it's _that_ kind of read-only.

But it might work, actually. Thanks! 

I'll send a PR if it works out

PR #287, opened 2020-07-24
Add read only attribute to test model

StefSchenkelaars:
I ran into the same problem as described in https://github.com/magnusvk/counter_culture/issues/278. First I've added a read only attribute to replicate the error in the test suite and then I changed the way the attributes are restored in the `previous_model` method. 

This way, it skips the overwritten setters which I think also makes sense to do since it's really the dirty changes as they will be persisted to the database. Aasm is also overwriting the setter and read only attributes are basically doing the same thing. I can't come up with a use case where this will not work but it's definitely good to think this trough. 

Closes https://github.com/magnusvk/counter_culture/issues/278

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -290,7 +290,7 @@ module CounterCulture
       changes_method = ACTIVE_RECORD_VERSION >= Gem::Version.new("5.1.0") ? :saved_changes : :changed_attributes
       obj.public_send(changes_method).each do |key, value|
         old_value = ACTIVE_RECORD_VERSION >= Gem::Version.new("5.1.0") ? value.first : value
-        prev.public_send("#{key}=", old_value)
+        prev[key] = old_value
       end
 
       prev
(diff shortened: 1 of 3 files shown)


lib/counter_culture/counter.rb
```diff
@@ -290,7 +290,7 @@ def previous_model(obj)
       changes_method = ACTIVE_RECORD_VERSION >= Gem::Version.new("5.1.0") ? :saved_changes : :changed_attributes
       obj.public_send(changes_method).each do |key, value|
         old_value = ACTIVE_RECORD_VERSION >= Gem::Version.new("5.1.0") ? value.first : value
-        prev.public_send("#{key}=", old_value)
+        prev[key] = old_value
```
magnusvk, line 293: Reading up the API documentation, honestly this looks to be the "more correct" way to do this anyhow. So if it also fixes your issue with aasm, all the better!

magnusvk:
Thank you, released as gem version 2.6.0.

Issue #289, opened 2020-08-08
Improving the use of column_names and scopes


maikdijkstra:
As of now the column column_name is always necessary to create a counter. With some things this would result in large pieces of code which could be minimized a lot using only the column_names. Especially when using scopes.

For example:
```ruby
class City < ActiveRecord::Base
  belongs_to :prefecture
  scope :big, -> { where('population > ?', 100000) 
  scope :small, -> { where('population < ?', 10000) }
  scope :small_and_big_cities, -> { where('population > ? OR population < ?', 100000, 10000) }

  counter_culture(
    :prefecture,
    column_name:  ->(model) { model.big? ? :big_cities_count : nil },
    column_names: { 
        City.big => :big_cities_count
    }
  )

counter_culture(
    :prefecture,
    column_name:  ->(model) { model.small? ? :small_cities_count: nil },
    column_names: { 
        City.small => :small_cities_count
    }
  )

  counter_culture(
    :prefecture,
    column_name:  ->(model) { (model.big? || model.small?) ? :small_and_big_cities_count : nil },
    column_names: { City.big => :small_and_big_cities_count }
  )

  def big?
     population > 100000
  end
  def small?
     population < 10000
  end
end
```
This is a small example with just 2 options and a combination of those. But takes quite a lot of lines. I would like to write it like this to easily create more counters without having large pieces code 
```ruby
class City < ActiveRecord::Base
  belongs_to :prefecture
  scope :big, -> { where('population > ?', 100000) }
  scope :small, -> { where('population < ?', 10000) }
  scope :small_and_big_cities, -> { where('population > ? OR population < ?', 100000, 10000) }

  counter_culture(
    :prefecture,
    column_names: { 
      City.big => :big_cities_count,
      City.small => :small_cities_count,
      City.small_and_big_cities => :small_and_big_cities_count
    }
  )
end
```
In this way the column_names is used as a kindof mapping. I have edited the code to handle this situation and I think this would be a very nice addition to this already good gem! Please let me know if you want me to open a PR for this.

magnusvk:
I agree it would be nice to be able to define your column names just once, I just don't know how to get there consistently and while supporting the full functionality of the gem. #290 I don't think gets us there. I'll close this here as I don't know how I'd achieve this, but I'm happy to keep discussing more ideas as I agree it would be nice to have this functionality.

PR #290, opened 2020-08-09
added functionality

maikdijkstra:
I added the possibilty to define only the column_names to easily create many counters using scopes. I know that the solution is not super neat, but that is because the previous_model is never part of the current scope so I had to calculate if the previous model would be part of the scope. This closes the issue #289 that I made. Most of the explanation is also in there what the code actually achieves

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -3,7 +3,7 @@ module CounterCulture
     CONFIG_OPTIONS = [ :column_names, :counter_cache_name, :delta_column, :foreign_key_values, :touch, :delta_magnitude]
     ACTIVE_RECORD_VERSION = Gem.loaded_specs["activerecord"].version
 
-    attr_reader :model, :relation, *CONFIG_OPTIONS
+    attr_reader :model, :relation, *CONFIG_OPTIONS, :without_column_name
 
     def initialize(model, relation, options)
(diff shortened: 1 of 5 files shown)


lib/counter_culture/counter.rb
```diff
@@ -130,6 +137,47 @@ def counter_cache_name_for(obj)
...
+      @column_names.each do |scope, column_name|
+        column_names << column_name if scope.exists?(obj.id) || obj.instance_eval(sql_query_to_eval_query(obj, scope.to_sql))
+      end
+      column_names
+    end
+
+    # Parse sql query with AND, OR, IN to a logical expression
+    #
+    # sql: the query to parse
+    def sql_query_to_eval_query(obj, sql)
```
magnusvk, line 154: I am not onboard with this approach. As you say, it's not neat. But more importantly a scope can be any set of valid SQL conditions—not just the ones you cover here. And given different databases there's really an unbounded set of allowable scope queries so I don't think it's even possible to cover them all here.

I think I understand the reason you are doing this here—to check the `previous_model` against the scope. The only other approach I can come up with here to cover that would be to actually save the `previous_model` to the database, check against it and then roll back. But I don't like that approach either. It assumes you actually have the ability to roll back, which is not always a given, plus it probably has terrible performance impacts.

PR #292, opened 2020-09-03
fixes last argument as keyword parameters deprecation warning

thdaraujo:
Thanks for creating this gem! It's super useful!

------------

On ruby 2.7.0, using the last argument as keyword parameters is deprecated (see [announcement](https://github.com/ruby/ruby/blob/4643bf5d55af6f79266dd67b69bb6eb4ff82029a/doc/NEWS-2.7.0#the-spec-of-keyword-arguments-is-changed-towards-30-)).

This PR fixes the following deprecation warning for ruby 2.7.1 on `lib/counter_culture/reconciler.rb:112`
```
counter_culture-2.6.0/lib/counter_culture/reconciler.rb:112: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

### DIFF
--- a/lib/counter_culture/reconciler.rb
+++ b/lib/counter_culture/reconciler.rb
@@ -109,7 +109,7 @@ module CounterCulture
           find_in_batches_args[:start] = options[:start] if options[:start].present?
           find_in_batches_args[:finish] = options[:finish] if options[:finish].present?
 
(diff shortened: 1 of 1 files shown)


magnusvk:
Thank you! Merged and released as gem version 2.6.1.

thdaraujo:
great, thank you! 💯

Issue #96, opened 2015-02-16
Manually populating counts doesn't work with Conditional counter cache


nbhartiya:
Manually populating counts (https://github.com/magnusvk/counter_culture#manually-populating-counter-cache-values) with Conditional Counter Cache (https://github.com/magnusvk/counter_culture#conditional-counter-cache) leads to error: 

RuntimeError: Must provide :column_names option for relation [:category] when :column_name is a Proc; you may skip this relation with :skip_unsupported => true

How should I get around this and actually manually populate counts for my conditional counter cache?

magnusvk:
This is well-documented behavior; please read the documentation before filing issues.
https://github.com/magnusvk/counter_culture/#handling-dynamic-column-names

magnusvk:
@bartoszkopinski this is an open-source project, you know. Feel free to open a PR to improve the documentation or the behavior itself. I'd be excited to look at it.

Also, frankly, people with your kind of attitude ^^ sometimes make me wonder why I give my time away for free here. So if you don't like my project, feel free to go away.

bartoszkopinski:
@magnusvk You were the one with the attitude criticising OP for not reading docs while the docs are literally wrong and incomplete. 
Same [here](https://github.com/magnusvk/counter_culture/issues/118) and [here](https://github.com/magnusvk/counter_culture/issues/130).

magnusvk:
¯\_(ツ)_/¯

I tried to write docs that explain this, to the best of my ability. If you have a better suggestion, stop complaining and open a PR. That is literally how open source works.

PR #294, opened 2020-09-16
Update document about Conditional counter cache

ykamez:
Thanks for creating this gem! It's super useful!

---

I ran into a problem referenced in https://github.com/magnusvk/counter_culture/issues/96, so I updated documentation.
Please take a look at it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,8 @@ end
 
 Now, the ```Category``` model will keep the counter cache in ```special_count``` up-to-date. Only products where ```special?``` returns true will affect the special_count.
 
+If you would like to use this with `counter_culture_fix_counts`, make sure to also provide [the `column_names` configuration](#handling-dynamic-column-names).
+
 ### Totaling instead of counting
 
 Instead of keeping a running count, you may want to automatically track a running total.


magnusvk:
Thank you—however, as it stands your change isn't quite accurate. Providing `column_names` is optional and only required if you want to use `counter_culture_fix_counts`. Also, there already exists documentation on this: https://github.com/magnusvk/counter_culture#handling-dynamic-column-names

Perhaps we can just provide a link to the ☝️ existing documentation, along the lines of "if you would like to use this with `counter_culture_fix_counts` make sure to also provide the `column_names` configuration." Would that have been helpful?

ykamez:
Thank you for your response!

> Thank you—however, as it stands your change isn't quite accurate. Providing `column_names` is optional and only required if you want to use `counter_culture_fix_counts`. Also, there already exists documentation on this: https://github.com/magnusvk/counter_culture#handling-dynamic-column-names


Right.

> Perhaps we can just provide a link to the ☝️ existing documentation, along the lines of "if you would like to use this with `counter_culture_fix_counts` make sure to also provide the `column_names` configuration." Would that have been helpful?

It sounds good. I'll add the lines and the link.

magnusvk (approved):
Thank you!

ykamez:
Thank you!

PR #295, opened 2020-09-21
[WIP] apply scope if available in polymorphic case

jgigault:
I am encountering an issue with polymorphic associations having a defined scope, as follows:

```ruby
class Asset
  belongs_to :assetable, -> { unscope(where: :state) }, polymorphic: true
  counter_culture :assetable
end

class Project
  default_scope { where(state: 1) }
  has_many :assets, as: :assetable
end
```

The resulting query of the gem to update counters does not take into account the `unscope` statement.
I have quickly committed this fix to get your feedback and maybe go further with a complete PR.

Thanks for your help!

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -205,11 +205,15 @@ module CounterCulture
         # this is the column that stores the polymorphic type, aka the class name
         type_column = reflect.foreign_type.to_sym
         # so now turn that into the class that we're looking for here
-        if was
-          attribute_was(source, type_column).try(:constantize)
-        else
-          source.public_send(type_column).try(:constantize)
-        end
+        new_klass =
+          if was
+            attribute_was(source, type_column).try(:constantize)
+          else
+            source.public_send(type_column).try(:constantize)
+          end
+        return new_klass if reflect.scope.nil?
+        # in case of a scope is made available, merge it
+        new_klass.merge(reflect.scope)
       else
         reflect.klass
       end


lib/counter_culture/counter.rb
```diff
@@ -205,11 +205,15 @@ def relation_klass(relation, source: nil, was: false)
...
+          if was
+            attribute_was(source, type_column).try(:constantize)
+          else
+            source.public_send(type_column).try(:constantize)
+          end
+        return new_klass if reflect.scope.nil?
+        # in case of a scope is made available, merge it
+        new_klass.merge(reflect.scope)
       else
         reflect.klass
```
jgigault, line 218: Maybe this case is also an issue, I will check if the proposed fix would need to be extended

magnusvk:
So, to be honest, my knee-jerk reaction is "don't use default scopes" because they really, truly are nothing but trouble, as of course you can see here. However, if there's a straightforward fix here somewhere I'm willing to consider it.

Two things though: I'd want to start with a failing test case (or multiple failing test cases potentially—this probably affects more than just polymorphic relationships, right?). Ideally have them in a separate commit so we can see they actually fail without the fix. And then I think the fix fits better in the `change_counter_cache` method where there's actually a query constructed than here where it's really just supposed to return a `Class`.

jgigault:
Dear @magnusvk 
This seems not to be only an issue related to default scope, but applied filters on associations.
Thanks for you reply, I will look deeper into `change_counter_cache` for the next iteration.

magnusvk:
Closing for inactivity.

PR #297, opened 2020-10-21
Remove dependency to after_commit_action gem

onk:
`execute_after_commit` option has been removed in v2.0.0.


magnusvk:
You are so right, thank you! Released as gem version 2.6.2.

PR #300, opened 2020-10-30
Add a column_name option for fixing counts 

lollkebear:
Add the support for the `column_name` option for `Model.counter_culture_fix_counts` to fix counts on the specific column. This allows us to skip the columns that have already been processed. This can be useful after running big DB changes that affect only one counter cache column.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,10 @@ Product.counter_culture_fix_counts only: :category
 Product.counter_culture_fix_counts only: [[:subcategory, :category]]
 # will automatically fix counts only on the two-level [:subcategory, :category] relation on Product
 
+Product.counter_culture_fix_counts column_name: :reviews_count
+# will automatically fix counts only on the :reviews_count column on Product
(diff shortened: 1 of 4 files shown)


magnusvk:
Thanks for the PR! It makes sense you'd want to fix only one column, but couldn't you use the existing `only` option for that? Maybe I'm missing something, but at first glance this looks to me like it's duplicating the `only` option.

lollkebear:
The `only` option doesn't allow us to change only one column. Instead, it changes all columns for one relation. 
Here we have a situation where we have 2 columns we can fix for the `Product` model.
```
  counter_culture :category,
      column_name: proc {|model| "#{model.product_type}_count" },
      column_names: {
          ["products.product_type = ?", 'awesome'] => 'awesome_count',
          ["products.product_type = ?", 'sucky'] => 'sucky_count'
      }
```
By using the `only` option it fixes both columns. However, with the `column_name` option we can fix either `awesome_count` or `sucky_count` separately

magnusvk:
That's a good point! Released as gem version 2.7.0. Thanks for contributing!

PR #303, opened 2021-01-08
Start testing Ruby 3.0 and Rails 6.1

magnusvk:


xinranxiao:
How did the merge go?

magnusvk:
No issues with the gem, it's just paper_trail that's having issues, so that integration isn't getting tested for Ruby 3.0 or Rails 6.1. Still waiting on those guys to merge the fix. But if you want to use the gem you shouldn't have an issue with any of those versions.

Issue #299, opened 2020-10-26
Delay / Defer UPDATE call


PhilCoggins:
Is there a way to delay counter updates until a later time? I'm thinking in the case where records are inserted in bulk (like imports), it would be nice to wrap the inserts in a block that would prevent UPDATEs from occurring, and then when the process is complete I could just call `counter_culture_fix_counts` myself. Yes, I could use a bulk insert tool, but there are certain callbacks and validations that I still wish to run.

I have not seen anything in Readme and a quick look at the repo shows this isn't currently possible. Would this library be interested in having this functionality added?

magnusvk:
This isn’t currently possible. My sense is this probably isn’t a super common use case, but if there’s a clean way to add this to the gem I would consider it. 

If this is just a one-off process for you and you’re ok with a hackish solution, the easiest way to accomplish this is probably to empty the internal config array and then repopulate it when you’re done: https://github.com/magnusvk/counter_culture/blob/fcf5a0971cbb68142b696038c187dc80f93db909/lib/counter_culture/extensions.rb#L53

PhilCoggins:
@magnusvk Thank you for the suggestion. Emptying `@after_commit_counter_cache` does not appear to be a threadsafe solution, so I'll do some digging, and open a PR if I can throw something reasonable together.

magnusvk:
Yeah, that sounds right. If you want this to be thread-safe I’d imagine you need to modify the counter culture code and set a flag on the thread.

PhilCoggins:
This is what I ended up with:
```
module CounterCulture
  module Extensions
    ClassMethods.module_eval do
      def defer_counter_culture_updates
        counter_culture_updates_was = Thread.current[:defer_counter_culture_updates]
        Thread.current[:defer_counter_culture_updates] = Array(counter_culture_updates_was) + [self]
        yield
      ensure
        Thread.current[:defer_counter_culture_updates] = counter_culture_updates_was
      end
    end
  end

  module DeferredUpdate
    private

    def _update_counts_after_create
      unless Array(Thread.current[:defer_counter_culture_updates]).include?(self.class)
        super
      end
    end

    # called by after_destroy callback
    def _update_counts_after_destroy
      unless Array(Thread.current[:defer_counter_culture_updates]).include?(self.class)
        super
      end
    end

    # called by after_update callback
    def _update_counts_after_update
      unless Array(Thread.current[:defer_counter_culture_updates]).include?(self.class)
        super
      end
    end
  end
[...]

PR #304, opened 2021-01-13
Add attr_accessor to skip counts after create

benjaminwols:
I have a counter which gets triggered a lot on create (+-100 per second) which causes some performance issues.
I wanted to disable counter_culture on create and manually fix the count at a later stage.
I saw this issue: https://github.com/magnusvk/counter_culture/issues/299, but decided to implement it in a different way.
Please let me know if you are willing to add this to the gem. If so I can also update the readme.

### DIFF
--- a/lib/counter_culture/extensions.rb
+++ b/lib/counter_culture/extensions.rb
@@ -2,6 +2,10 @@ module CounterCulture
   module Extensions
     extend ActiveSupport::Concern
 
+    included do
+      attr_accessor :skip_counts_after_create
+    end
+
     module ClassMethods
       # this holds all configuration data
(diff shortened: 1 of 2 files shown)


magnusvk:
Two thoughts in response here:
1. Why just the `after_create` call? Isn't this a more general issue that might also apply for updates? If we were to include this in the gem, I would probably want to allow disabling count updates completely.
2. Why is this a flag on the instance level? Wouldn't it be more convenient to just wrap a whole block of code and say "don't update any counts on the model `SomeModel` in this block?"

magnusvk:
Closing this for lack of response. Happy to re-evaluate with more discussion.

PR #310, opened 2021-03-09
Enable bundler caching for travis

Taher-Ghaleb:
Would be interested to know why bundler caching of Travis hasn't been enabled in this repository. Thank you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+cache: bundler
 services:
   - postgresql
   - mysql


magnusvk:
Just never occurred to me. I wonder if this is working, though? I restarted one of the jobs in this build and it didn't use the cache. Any way to verify this is actually working?

Taher-Ghaleb:
Oh, I see. I checked on my side and it appears that the cache size is zero (see below). At the end of the build log, it says _changes detected_ and then _cache uploaded_, but it's still empty. Not sure to be honest. Maybe path issues?


image

magnusvk:
Yeah probably a path issue. I'm going to close this for now as I don't have the time to figure out what's going on. But if you want to tinker with it and can get it to work I'm definitely down to turn this on. Thanks!

Taher-Ghaleb:
Fixed by [#311](https://github.com/magnusvk/counter_culture/pull/311).

Issue #312, opened 2021-03-13
counter_culture_fix_counts not working on model with default_scope defined


xhs345:
I have two models:

```
Class User
  default_scope {includes(:location)}

  has_many :sleeps

  # major_sleeps_count
...
end
```

```

Class Sleep
  belongs_to :user
  counter_culture :user,
                  column_name: proc { |model| model.major_sleep.present? ? 'major_sleeps_count' : nil },
                  column_names: { ["sleeps.major_sleep IS NOT NULL"] => 'major_sleeps_count' }
  ...
end
```

When running `Sleep.counter_culture_fix_counts` I'm getting the error:

> ActiveModel::MissingAttributeError (missing attribute: location_id)

magnusvk:
So I have to admit I don't 100% understand why this is getting raised, but adding an `includes` on a default scope is almost certainly a bad idea. I would advise you to drop that and add a scope instead that you can use everywhere you want to add that `includes`.

xhs345:
Oh, I totally agree, but sadly that is not easily done in my case since it's an existing application

PR #313, opened 2021-03-13
Unscope relation class by default

xhs345:
Fixes issue #312 by removing default scopes from class by default

### DIFF
--- a/lib/counter_culture/reconciler.rb
+++ b/lib/counter_culture/reconciler.rb
@@ -91,7 +91,7 @@ module CounterCulture
           relation_class_table_name = quote_table_name(relation_class.table_name)
 
           # select join column and count (from above) as well as cache column ('column_name') for later comparison
-          counts_query = scope.select(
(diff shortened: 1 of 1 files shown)


magnusvk:
I don't think this is the right approach. For example, this will fix counts on all soft-deleted records. My knee-jerk reaction is that your problem is that you're using default scopes and that's usually a bad idea and you should stop. 😉 The maybe-more-helpful version is to say: Could we add an option here that lets you pass in the scope that you want to use when fixing counts? Then you can pass in `unscoped` there and get this behavior without changing behavior for everybody.

xhs345:
Ok, didn't think about the soft deletes. 
As mentioned in the other comment: removing the default_scope from the model is not an option at the moment sadly. I'll look into passing the scope as a parameter

magnusvk:
Gotta love those legacy applications. Thanks for taking a look at that—and just a note that we should also have a test for that behavior.

PR #326, opened 2021-07-06
Added possibility to pass Procs to execute_after_commit

marinfr:
This is useful to temporarily allow real-time counter updates in a controlled scope.

Example:

```ruby
  # models/comment.rb
  counter_culture :post, execute_after_commit: proc { !Thread.current[:realtime_counter] }
```

```ruby
  Post.transaction do
    post = Post.lock.last
    with_realtime_counters do
      post.comments.last.destroy!
    end
  end
  
  def with_realtime_counters(&block)
    Thread.current[:realtime_counter] = true
    yield
  ensure
    Thread.current[:realtime_counter] = nil
  end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,12 @@ Another option is to simply defer the update of counter caches to outside of the
   counter_culture :category, execute_after_commit: true
 ```
 
+You can also pass a `Proc` for dynamic control. This is useful for temporarily enabling realtime counters:
+
+```ruby
+  counter_culture :category, execute_after_commit: proc { !Thread.current[:realtime_counter] }
+```
+
 ### Manually populating counter cache values
 
 You will sometimes want to populate counter-cache values from primary data. This is required when adding counter-caches to existing data. It is also recommended to run this regularly (at BestVendor, we run it once a week) to catch any incorrect values in the counter caches.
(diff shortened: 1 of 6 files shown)


magnusvk:
@marinfr what's your use case for this? Even with `execute_after_commit` set to true counters are still "real-time," right? The only question is if they happen just before or just after the commit. 

This looks to me like it adds a bunch of complication and I don't understand the benefit, so I'm inclined not to merge this unless I'm missing something here.

marinfr:
@magnusvk we had some application-wide problems with `LockWaitTimeout` errors, so we added the `execute_after_commit: true` option. In my case, it's ok for most counters to execute after commit. There are, however, specific cases where I need to decide whether to commit or roll back a big transaction based on the counter's value - which is impossible with `execute_after_commit: true` since the counter is not updated inside the transaction. The only solution would be to execute a count query, which defeats the purpose of counter caching. But passing a `proc` (similar to `counter_name`) allows me to enable real-time counters only for and during these specific transactions.

marinfr:
Of course, my description and example with thread variables are based on a specific issue, but I think dynamic `execute_after_commit` could be used in other ways as well!

PR #330, opened 2021-10-08
Support for db replica when using counter_culture_fix_counts

cameronmccord2:
Add support for sending fix count queries to a database replica.

```ruby
Product.counter_culture_fix_counts db_connection_builder: proc{|reading, block|
  if reading # Count calls will request a reading connection
    Product.connected_to(role: :reading, &block)
  else # Update all calls will request a non-reading connection
    Product.connected_to(role: :writing, &block)
  end
}
```

Story Reasoning:

Every so often we recalculate all of our counter culture counts using Model.counter_culture_fix_counts. Our 'need fixed' counts are spread through our table so we have to iterate the whole thing instead of just the 'newest' records at the end. Our table with 20 million rows and 1 counter cache issues 20,000 count queries(batch_size: 1000) and only about 5% of our count queries result in an Update All database hit. The recount is done in the low traffic time and split across multiple jobs but we'd rather not pound our primary database. We have 22 counters spread across 4 models.
We've been using this in production for a few weeks with no issues.

Implementation Reasoning:

Allowing the user to define a proc that can wrap our database call means they can implement it in any way they want using any library they want. Rails 6 native support uses this construct `Product.connected_to(role: :reading, &block)` but there are numerous options that can be passed into this function. Having counter culture call this method directly with the `:reading`, `:writing`, init specified, or function arg specified role wouldn't account for other options or other patterns.


spec/counter_culture_spec.rb
```diff
@@ -1441,6 +1441,59 @@
...
+      if reading
+        requested_reading_connection = true
+      else
+        requested_writing_connection = true
+      end
+      block.call
+    }
+    expect(requested_reading_connection).to be(true)
+    expect(requested_writing_connection).to be(true)
+  end
```
magnusvk, line 1495: Is there a way to have a separate read-only DB user in the test here? If so it would be really nice to also have a test that actually passes the read-only connection here to make sure that the write happens on the write connection as it should.

magnusvk (approved):
This looks good, sorry for the delay here. Can you resolve the merge conflicts?

magnusvk:

Issue #298, opened 2020-10-24
column_names option lead to database not exist when run rake db:create


SmileDMiao:
Rails: 6.0
counter_culture: 2.6.2

I found if i delete the column names option the `rake db:create` command will success. when add it back , run `rake db:create` raise database not exist error.

```ruby
counter_culture :user, column_name: proc { |order| order.checked? ? 'spent_amount' : nil }, delta_column: 'spent', column_names: {
      Prediction::Order.checked => 'spent_amount',
    }
```

SmileDMiao:
change the code
```ruby
 counter_culture :user, column_name: proc { |order| order.checked? ? 'spent_amount' : nil }, delta_column: 'spent', column_names: {
      ["prediction_orders.status = ?", 2] => 'spent_amount',
    }
```
will not raise the error.

magnusvk:
Looking at this, what's probably going on is that `rake` is loading your app on `db:create` which triggers the `Prediction::Order.checked` code which will in turn raise the "database does not exist" error. I'm not really sure how to work around this, to be honest, other than what you've already done here. I'm going to close this as this isn't strictly related to counter_culture, but if you have a good idea on how to work around this I'm happy to discuss it.

Archetylator:
Probably to resolve that issue you can put counter definitions inside this code:

```
if Prediction::Order.table_exists?
...
end
```

magnusvk:
That could work, feel free to open a PR for it and I'm happy to take a look!

PR #335, opened 2021-11-24
Allows providing a Proc to column_names.

PhilCoggins:
* Returning a direct hash with a scope triggers loading of the schema cache in ActiveRecord, which can be undesirable
when your app is being booted. Using a Proc instead defers hitting this scope until it is actually called.

Fixes https://github.com/magnusvk/counter_culture/issues/298

It is discouraged to hit your database while code is being loaded during the boot process, this can affect such tasks as `db:create`, or building a Docker image that doesn't have access to a database. We can just use a Proc so the scope isn't called until it's actually needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,7 +354,9 @@ class Product < ActiveRecord::Base
 end
 ```
 
-You can specify a scope instead of a where condition string for `column_names`:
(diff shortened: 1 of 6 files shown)


magnusvk:
Thank you @PhilCoggins — I like this approach a lot! Released as 3.1.0.

PR #336, opened 2021-12-28
Allow specifying the polymorphic_classes when calling counter_culture_fix_counts

kleinjm:
### What

Provides the option of specifying the `polymorphic_classes` when calling `counter_culture_fix_counts`

### Why

I discovered that the `DISTINCT` query used in polymorphic relationships could be very expensive on large tables. The screenshot below shows that the majority of time spent in order creation on our marketplace platform is from this query.

image

The result of the query is the name of the classes to be updated. It's common in our use cases, [here at Patch](https://www.patch.io/), that we know the associated model or models that need to be updated. Therefore, providing this option will eliminate the potentially expensive query. All existing behavior is preserved.

### Notes

This change is live in production on our fork of the repo. I used [db-query-matchers](https://github.com/civiccc/db-query-matchers) to confirm the fix when testing locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,7 +378,7 @@ end
 
 You can specify a scope instead of a where condition string for `column_names`. We recommend
 providing a Proc that returns a hash instead of directly providing a hash: If you were to directly
-provide a scope this would load your schema cache on startup which will break things like 
+provide a scope this would load your schema cache on startup which will break things like
 `rake db:migrate`.
 
 ```ruby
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -459,6 +459,17 @@ end
...
 
+To discover which models need to be updated via `counter_culture_fix_counts`,
+counter_culture performs a `DISTINCT` query on the polymorphic relationship.
+This query can be expensive so we therefore offer the option
+(`polymorphic_classes`) to specify the models' counts that should be corrected:
+
+```ruby
+Image.counter_culture_fix_counts(polymorphic_classes: Product)
+# or
+Image.counter_culture_fix_counts(polymorphic_classes: [Product, Employee])
```
pcothenet, line 470: Note: I worked with @kleinjm on this issue. It would be neat if this could be inferred from the Rails API, but we haven't found an easy way. Open to suggestions!

magnusvk:
Thanks for the contribution! This makes sense to me as an optimization. I'm not too worried about not inferring this automatically given that it's optional.

magnusvk:
Released as gem version 3.2.0. Thanks for your contribution!

PR #337, opened 2022-01-03
Fix primary key name for enumerable relations

samirph:
This fixes an issue that happens when you have an array of relations, such as `[:sub_group, :group]` and their id keys are different. For example, one id key is `id` and another one is `uuid` or `uid` or anything else.

Before this fix, the method would `try` to access the last relation id using a previous relation key.


magnusvk:
This looks good, thank you! Just added one more spec to make sure that `counter_culture_fix_counts` works as well. This passes locally, will merge when that goes green.

magnusvk:
Released as gem version 3.2.0. Thanks for your contribution!

PR #338, opened 2022-01-04
Fix primary key name for enumerable relations

samirph:
This fixes an issue that happens when you have an array of relations, such as [:sub_group, :group] and their id keys are different. For example, one id key is id and another one is uuid or uid or anything else.

Before this fix, the method would try to access the last relation id using a previous relation key.


magnusvk:
👋🏻  thanks for contributing, wanted to check in why you closed this. Is the issue resolved or did you have trouble getting the tests green?

PR #339, opened 2022-01-25
Test Rails 7.0 and latest Ruby patch versions

magnusvk:
Also clean up `Appraisals` now that we've dropped tests against older
Rails versions.


magnusvk:
Already merged as part of #340.

PR #343, opened 2022-02-23
Add Ruby 3.1 to CI

petergoldstein:
Adds Ruby 3.1 to the CI matrix.  I used an include approach as that required the least change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,16 @@ workflows:
               ruby-version: ["2.6", "2.7", "3.0"]
               rails-version: ["5.2", "6.0", "6.1", "7.0"]
               database: ["postgresql", "sqlite3", "mysql2"]
+            include:
+              - ruby-version: "3.1"
+                rails-version: "7.0"
+                database: "postgresql"
+              - ruby-version: "3.1"
+                rails-version: "7.0"
+                database: "sqlite3"
+              - ruby-version: "3.1"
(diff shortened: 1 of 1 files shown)


magnusvk:
Thank you—why only test against Rails 7 though? [Sounds as if Rails 6.1 should also be compatible.](https://discuss.rubyonrails.org/t/is-rails-6-1-x-compatible-with-ruby-3-1/79713/6)

petergoldstein:
@magnusvk There's some debate about whether Rails 6.1 is compatible with Ruby 3.1.  But most things seem to work.

That said, to get it to work generally requires a modification to the Rails 6.1 Gemfile.  We need to conditionally include the `net/smtp', 'net/pop', and 'net/imap' gems when the Ruby version is 3.1 or greater.   If you want me to add that let me know and I can make the necessary modifications.

petergoldstein:
@magnusvk Please let me know how you want to resolve the above comment.  I can add Rails 6.1 with the Ruby 3.1+ hackery in the rails_6_1.gemfile or just leave it as is.  And if there's anything else required to get this merged, please let me know.

petergoldstein:
@magnusvk Anything I can do to move this forward?  Thanks.

magnusvk:
Sorry for the radio silence, I just tried this in #357 and it worked fine without extra changes. Thanks for the contribution!

PR #346, opened 2022-03-04
Differentiate installation steps for current versus legacy ruby/rails versions

edaff-iti:


magnusvk:
Thanks, but since those versions are all end of life I don't think it's worth pointing this out. I'll update the readme to suggest installing the latest version instead.

PR #351, opened 2022-04-20
Fix issue that decrement counter cache is called multiple times

kmdtmyk:
For example, there are these models.

```ruby
class Book < ApplicationRecord
  has_many :reviews
end

class Review < ApplicationRecord
  belongs_to :book
  counter_culture :book
end
```

I called review.destory method twice. I expected this decrements counter -1. But, this decrements counter -2 acutually.
I fixed it.

```ruby
book = Book.create
review = Review.create(book: book)

book.reload
expect(book.reviews_count).to eq 1

review.destroy
book.reload
expect(book.reviews_count).to eq 0

review.destroy
book.reload
expect(book.reviews_count).to eq 0 # actually, -1 
```

### DIFF
--- a/lib/counter_culture/extensions.rb
+++ b/lib/counter_culture/extensions.rb
@@ -113,8 +113,10 @@ module CounterCulture
     # called by after_destroy callback
     def _update_counts_after_destroy
       self.class.after_commit_counter_cache.each do |counter|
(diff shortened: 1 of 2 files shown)


magnusvk:
Thank you!

magnusvk:

PR #352, opened 2022-04-25
Fix passing a proc to `column_names`

KevinCarterDev:
The documentation states

> You can specify a scope instead of a where condition string for `column_names`. We recommend
providing a Proc that returns a hash instead of directly providing a hash: If you were to directly
provide a scope this would load your schema cache on startup which will break things like
`rake db:migrate`.

This doesn't work tho as the proc is immediately `call`-ed in the `counter_culture` method to check it's return value. To fix this I've moved the ArgumentError from the setup side to the runtime side.

### DIFF
--- a/lib/counter_culture/extensions.rb
+++ b/lib/counter_culture/extensions.rb
@@ -45,15 +45,7 @@ module CounterCulture
           @after_commit_counter_cache = []
         end
 
-        column_names_valid = (
-          !options[:column_names] ||
-          options[:column_names].is_a?(Hash) ||
-          (
-            options[:column_names].is_a?(Proc) &&
-            options[:column_names].call.is_a?(Hash)
-          )
-        )
-        unless column_names_valid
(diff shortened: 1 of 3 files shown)


AntoineBecquet:
`options[:column_names]` could still be checked as a Hash or a Proc on the setup side, as it was before this commit :
https://github.com/magnusvk/counter_culture/commit/2a86dd82daefdc75744d08b42e50a5681239ebb2#diff-34fcb2d6df6a065b09e89f5e0980d73d3bd56f9e91c1bd71ded2ff36610df30aL48-R60

But yes, the proc should not be called on startup.

KevinCarterDev:
I've updated it to add back in the check for a hash or proc and raise an error if it's not one of those.

nzifnab:
I'm running into this issue as well, any chance of getting this merged into the gem...? @magnusvk

magnusvk:

PR #353, opened 2022-06-01
Simplify fk lookup

gjtorikian:
Hiya @magnusvk 👋 Thanks for this great gem.

In my project, my Postgres database uses [binary ULIDs](https://github.com/rafaelsales/ulid)—they deserialize as strings, but are stored as binary data.

I noticed that my foreign key counters stopped working, and I tracked it down to the one line change in this PR. There are a couple of problems with the `"#{klass.table_name}.#{relation_primary_key(first, source: obj, was: was)} = ?"` format, the biggest being that the value is not [quoted](https://github.com/rails/rails/blob/de7c495209811f8f918fa9f915e64df61a6080c1/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L9-L10), which can cause security problems. Instead, we can lean heavily into `where` figuring out all the details for communicating with the database. 

While grepping around, I found two other places ([1](https://github.com/gjtorikian/counter_culture/blob/ffd172062876180ecbf7881cb7d6816f9ec23bbd/lib/counter_culture/counter.rb#L99), [2](https://github.com/gjtorikian/counter_culture/blob/ffd172062876180ecbf7881cb7d6816f9ec23bbd/lib/counter_culture/counter.rb#L122)) where this exact pattern is used, rather than `#{string} = ?`. That plus the passing tests gives me confidence that this is a safe change to make.

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -165,15 +165,13 @@ module CounterCulture
     def foreign_key_value(obj, relation, was = false)
       original_relation = relation
       relation = relation.is_a?(Enumerable) ? relation.dup : [relation]
-      
+
       if was
         first = relation.shift
         foreign_key_value = attribute_was(obj, relation_foreign_key(first))
         klass = relation_klass(first, source: obj, was: was)
         if foreign_key_value
-          value = klass.where(
-            "#{klass.table_name}.#{relation_primary_key(first, source: obj, was: was)} = ?",
-            foreign_key_value).first
+          value = klass.where(relation_primary_key(first, source: obj, was: was) => foreign_key_value).first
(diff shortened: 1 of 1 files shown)


gjtorikian:
@magnusvk Is this project still maintained?

magnusvk:
👋🏻 sorry for the radio silence, Github notifications let me down.

I agree that this change should be fine. In fact, though, I believe it doesn't change behavior at all. I disagree with your assertion that values aren't quoted, `where("something = ?", some_value)` _does_ quite the value if required. I don't mind merging this, per se, but if this _is_ changing behavior then I'd love to see a failing test case first. Do you think that's feasible to provide?

Issue #358, opened 2022-11-14
How to update with habtm when products have conditions?


exocode:
Hi there, 

I've some troubles get your gem running in my app:

Product has_and_belongs_to_many Categories through CategoriesProducts

But only "visible" products should be counted.

This is what I've so far (straight forward):

```ruby
class Product < ApplicationRecord
  scope :visible, -> { where(visible: true) }

  has_many :categories_products
  has_many :categories, through: :categories_products, class_name: 'Category'
end

class Category < ApplicationRecord
  has_many :categories_products
  has_many :products, through: :categories_products, class_name: 'Product'

  has_many :category_children, class_name: "Category", foreign_key: :category_parent_id
  belongs_to :category_parent, class_name: "Category", optional: true
end


class CategoriesProduct < ApplicationRecord
  belongs_to :category #, class_name: 'Category' 
  belongs_to :product #, class_name: 'Product'

  counter_culture :category,
                  # column_name: 'products_count', # works but does not filter my products by "visibility"
                  column_name: ->(model) { model.product.visible? ? 'products_count' : nil }
                  column_names: {
                    ['categories_products.visible', true] => 'products_count',
                  }
                  # also tried , but none of them are working.
                  column_names: {
                    Product.visible => 'products_count',
                  }
end

```

Do I have some extra scopes/queries in the relation model? 

Totally hung up. A point into the right direction whould save my day

Thank you

magnusvk:
Nothing looks wrong at first glance. Are you inserting `CategoriesProduct` without callbacks in some way perhaps? In that case counter_culture wouldn't work.

jwilson114:
Ive got an almost identical set up configuration and  counter_culture using column_names and a specific scope is not changing when the product_category relation is destroyed or created

`
counter_culture :product_category, column_name: 'active_products_count', execute_after_commit: true, column_names: -> { {
    ['products.active_sales_channel = TRUE'] => :active_products_count
  } }
`

would also appreciate any assistance

exocode:
> Nothing looks wrong at first glance. Are you inserting CategoriesProduct without callbacks in some way perhaps? In that case counter_culture wouldn't work.

Nothing special: 
`Product.update(categories: Category.all.sample(2)`
The new relations in `CategoriesProduct` are created, the old ones deleted. The new entries are properly counted, but the counter does not update the old entries.

When I do a `Product.unscoped.destroy_all` the `products_count` in the `Category` model wont get updated. ONLY a 
[...]

PR #359, opened 2022-11-17
added example when removed from habtm relation

exocode:
I've added rspec example for a habtm relation. 

- Given a `product.create("A product", categories: [old_category])`
- When `product.update(categories: [new_category1, new_category2])` is updating,
- The `old_category` relation is deleted, but the counter for `old_category` stays untouched. 

(The `new_categories` are counted properly!)

The other spec tests are mostly sanity checks

### DIFF
--- a/spec/counter_culture_spec.rb
+++ b/spec/counter_culture_spec.rb
@@ -34,6 +34,9 @@ require 'models/city'
 require 'models/group'
 require 'models/sub_group'
 require 'models/group_item'
+require 'models/habtm_product'
+require 'models/habtm_category'
+require 'models/habtm_category_product'
 
 if ENV['DB'] == 'postgresql'
   require 'models/purchase_order'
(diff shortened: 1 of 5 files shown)


exocode:
Regarding to this issue https://github.com/magnusvk/counter_culture/issues/358

magnusvk:
Ok, so I think the issue here must be that the callbacks on the has-and-belongs-to-many model aren't firing when you're updating the relation directly. Doing a little bit of googling on this, it seems like the way to do this would be to add `after_remove` and `after_add` callbacks on the relation. But then I'd imagine it getting tricky in that we'd have to make sure that only one or the other callback fires, not both, to prevent double-decrementing or -incrementing.

This unfortunately looks like it might end up being a pretty big change, so I won't have time to work on that. If you're willing to invest the time, I'd be happy to look at a PR that adds this. Alternatively, I think if you stopped using `update` and manipulated the has-and-belongs-to-many model directly instead, that it would end up counting correctly for you.

Thanks for opening this, it does make the problem easy to understand.

PR #362, opened 2023-01-01
Adds Ruby 3.2 to the CI matrix

petergoldstein:
I'm not sure if Ruby 3.2 will work with Rails 6.0 or 6.1. One or both of those may need to be excluded.


magnusvk:
Looks like it worked without issues. Thank you!

PR #367, opened 2023-04-01
Add support for multilevel many-to-many counters

SlayerDF:


lib/counter_culture/counter.rb
```diff
@@ -182,7 +182,11 @@ def foreign_key_value(obj, relation, was = false)
         value = value.send(relation.shift)
       end
 
-      return value.try(relation_primary_key(original_relation, source: obj, was: was).try(:to_sym))
+      if value.is_a?(ActiveRecord::Associations::CollectionProxy)
+        value.map { |v| v.try(relation_primary_key(original_relation, source: obj, was: was).try(:to_sym)) }
+      else
+        value.try(relation_primary_key(original_relation, source: obj, was: was).try(:to_sym))
```
magnusvk, line 188: Can we pull this out into a helper method like
```
def relation_primary_key_from_model(model, original_relation, obj, was)
  model.try(relation_primary_key(original_relation, source: obj, was: was).try(:to_sym)
end
```
Not sure if that's worth it, but it makes it clearer that the same thing is happening in both cases, just mapping in one.

spec/models/many_to_many/reader.rb
```diff
@@ -0,0 +1,4 @@
+class Reader < ActiveRecord::Base
```
magnusvk, line 1: I might be reading this wrong, but it looks like the `Reader` class isn't actually required for the test scenario here, is it?

magnusvk (commented):
I think this looks good—is it worth adding a note to the Readme file about this?

spec/models/many_to_many/reader.rb
SlayerDF: Actually, the problem with this scenario cannot be solved so easily, and the implementation is too naive. In the scenario, we have a `ReadersArticle` that updates the `readers_count` counter. Everything works fine until we delete the associated `AuthorsArticle`, which leads to the counter becoming out of sync.

Therefore, I don't see much benefit in this functionality. Unless leaving the technical possibility of specifying a multi-level path for the counter, and letting the implementation be up to a user, as it should be quite complex.

magnusvk: Should we close this then?

Issue #285, opened 2020-06-11
Supported Ruby version is still >= 2.3, right?


deanmi:
Hi,

First, I want to thank you for creating such an awesome gem! I'd like to make sure that I understand the change log comment for v2.5.0, which says dropped support for Ruby 2.3 and 2.4. The [gemspec still says 2.3 and greater](https://github.com/magnusvk/counter_culture/blob/78359792a287fbfe063609bfa2d558a289b37b8c/counter_culture.gemspec#L16), and looking at the commits for 2.5.0 it looks like you just dropped test support for 2.3 and 2.4.  Is that correct?  In other words, counter_culture should still work with Ruby 2.3 and above, correct?   Thanks!

magnusvk:
@deanmi well—it *probably* works. Given that it's not tested anymore and that Ruby 2.3 is EOL there's no guarantees. And I do remember there being at least one test failure—otherwise I wouldn't have dropped testing for it. But it's quite possible it was something obscure like a gemset failure or a feature you don't use. So yeah, if it works for you, no reason to panic. At the same time, you should probably upgrade anyhow. ;)

PR #369, opened 2023-06-08
Set required_ruby_version to officially supported Ruby versions

r7kamura:
As a user of this gem, when I run `bundle install`, I want it resolved to the latest version that is guaranteed to work with the Ruby version I am currently using.

From looking at CHANGELOG.md, it seems that this Gem currently only officially supports Ruby 2.6 and above. This leads me to believe that it would be better to keep the required_ruby_version in line with that as well.

What do you think?


r7kamura:
Maybe related issue:

- https://github.com/magnusvk/counter_culture/issues/285

PR #370, opened 2023-06-19
Handle reading primary key option from polymorphic association

jorge-d:
## Content

Allow reading the `primary_key` option on polymorphic associations when it's present instead of returning the association's `primary_key`.

Basically the issue is that I have an association defined such as:

```
  belongs_to :target, polymorphic: true, foreign_key: :target_uid, primary_key: :uid

  has_many :something_plurals, 
      as: :target,
      class_name: 'SomeClass',
      inverse_of: :target,
      foreign_key: :target_uid,
      primary_key: :uid
```

and yet `counter_culture` gem was trying to hit `Target#id` because `uid` is not the primary column of the table for now.

So my fix is to bypass the reflection on primary key if it's explicitly part of the polymorphic definition.


magnusvk (approved):
Looks good, thank you!

magnusvk:
Released with gem version 3.3.1.

PR #371, opened 2023-06-27
Skip counter culture update

RuslanHassonov:
Added the option to wrap code in `Model.skip_counter_culture_updates { ... }` block. With this block, we are able to skip counter_culture create, update and destroy callbacks when required and prevent unnecessary database queries.


magnusvk:
Could you just skip all callbacks instead? That strikes me as possibly the simpler solution here...

RuslanHassonov:
@benjaminwols can you please review this?

lib/counter_culture/extensions.rb
```diff
@@ -91,6 +91,14 @@ def counter_culture_fix_counts(options = {})
           reconciler.changes
         end.compact
       end
+
+      def skip_counter_culture_updates
+        counter_culture_updates_was = Thread.current[:skip_counter_culture_updates]
```
benjaminwols, line 96: Something like this
```suggestion
        return unless block_given?

        counter_culture_updates_was = Thread.current[:skip_counter_culture_updates]
```

benjaminwols (changes requested):

RuslanHassonov:
> Could you just skip all callbacks instead? That strikes me as possibly the simpler solution here...

@magnusvk I'm not entirely sure what you mean. Could you please make a suggestion?

magnusvk:
counter_culture uses ActiveRecord callbacks. So if you save / insert your records in a way that doesn't trigger callbacks (for example `insert_all` or `update_columns`) then counter_culture won't be triggered. And, generally speaking, if your updates are performance-sensitive then that's probably what you want to do. So I'm inclined to say it's not worth introducing this complexity in the gem.

RuslanHassonov:
The reason why I've developed this change is because I still want to run validations and some callbacks of our own, but not counter_culture callbacks (not until later, at least). `.insert_all` or `.update_columns` will skip everything, which I'm trying to avoid.

spec/counter_culture_spec.rb
```diff
@@ -98,6 +98,67 @@ def yaml_load(yaml)
...
+    user = User.create
+    product = Product.create
+
+    expect(user.reviews_count).to eq(0)
+    expect(product.reviews_count).to eq(0)
+    expect(user.review_approvals_count).to eq(0)
+
+    Review.skip_counter_culture_updates do
+      user.reviews.create :user_id => user.id, :product_id => product.id, :approvals => 13
+    end
```
magnusvk, line 111: Can you add some code below that makes sure the updates resume at the end of the block?

spec/counter_culture_spec.rb
```diff
@@ -98,6 +98,67 @@ def yaml_load(yaml)
...
+
+  it "skips increments counter cache on update" do
+    user = User.create
+    user.reviews.create :user_id => user.id, :approvals => 13
+
+    user.reload
+
+    expect(user.review_approvals_count).to eq(13)
+
+    Review.skip_counter_culture_updates do
```
magnusvk, line 129: Same here

spec/counter_culture_spec.rb
```diff
@@ -98,6 +98,67 @@ def yaml_load(yaml)
...
+
+    user.reload
+    product.reload
+
+    expect(user.reviews_count).to eq(1)
+    expect(product.reviews_count).to eq(1)
+    expect(user.review_approvals_count).to eq(13)
+
+    Review.skip_counter_culture_updates do
+      user.reviews.last.destroy
```
magnusvk, line 151: And here

spec/counter_culture_spec.rb
```diff
@@ -98,6 +98,67 @@ def yaml_load(yaml)
...
+    end
+
+    user.reload
+    product.reload
+
+    expect(user.reviews_count).to eq(1)
+    expect(product.reviews_count).to eq(1)
+    expect(user.review_approvals_count).to eq(13)
+  end
```
magnusvk, line 160: Can you also add a separate set of tests that makes sure this behavior works correctly when using nested calls to `skip_counter_culture_updates` across multiple models?

magnusvk (commented):
Fair enough. I'm frankly pretty reluctant to include this since it has never come up before, but it's simple enough that it's probably fine.

Can you also add some documentation for this in the readme file? Thanks.

RuslanHassonov:
@magnusvk Apologies for the late reply, I'll apply the suggested changes as soon as I'm able to.

magnusvk (approved):
Thank you!

magnusvk:
Released as gem version 3.4.0

PhilCoggins:
> Fair enough. I'm frankly pretty reluctant to include this since it has never come up before, but it's simple enough that it's probably fine.

*https://github.com/magnusvk/counter_culture/issues/299 has entered the chat*

PR #375, opened 2023-08-24
Add options as an argument for column_names proc

vsevolod:
Hello. We are using column_names method for recalculating total counters. And found a problem with the large SQL scope for update:
```ruby
class Attendance < ApplicationRecord
  ...
  counter_culture :event,
                  column_name: :missing_attendees_count,
                  column_names: -> { {Attendance.can_not_participate => :missing_attendees_count} }
```

```ruby
ids_for_update = {id: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}
Attendance.counter_culture_fix_counts(where: { Event.table_name => ids_for_update })
```

The current logic builds an SQL query like this:
```sql
SELECT "events".id, "events".id,
  COUNT("attendances".id)*1 AS count,
  MAX("events".missing_attendees_count) AS missing_attendees_count
FROM "events"
LEFT JOIN "attendances" AS attendances ON "events".id = attendances.event_id AND attendances.id IN (
  /*<<LARGE SELECT SQL SCOPE>>*/
)
WHERE "events"."id" IN (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
GROUP BY "events"."id"
ORDER BY "events"."id" ASC
LIMIT 1000
```
To get the data, the database should first load all the nested SQL.
I suggest pushing options to column_names for further filtering within the scope if needed:
```ruby
  counter_culture :event,
                  column_name: :missing_attendees_count,
                  column_names: -> (options) { {Attendance.can_not_participate(options) => :missing_attendees_count} }
```

There is only one problem with backward compatibility, if someone uses lambda for column_names instead of proc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,6 +431,23 @@ dynamic, you can pass `skip_unsupported`:
 Product.counter_culture_fix_counts skip_unsupported: true
 ```
 
+You can also use context within the block that was provided with the `column_names` method:
+
+```ruby
+class Product < ActiveRecord::Base
+  belongs_to :category
+  scope :awesomes, -> (ids) { where(ids: ids, product_type: 'awesome') }
+
+  counter_culture :category,
+      column_name: 'awesome_count'
+      column_names: -> (context) {
+        { Product.awesomes(context[:ids]) => :awesome_count }
+      }
+end
+
+Product.counter_culture_fix_counts(context: { ids: [1, 2] })
+```
+
 #### Handling over-written, dynamic foreign keys
 
 Manually populating counter caches with dynamically over-written foreign keys (```:foreign_key_values``` option) is not supported. You will have to write code to handle this case yourself.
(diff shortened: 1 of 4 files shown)


magnusvk:
What's the error if someone uses a lambda here? Will it fail because the lambda doesn't accept arguments? Would it be possible to check this and only pass the options if the lambda _does_ accept it as an argument? I don't feel good about merging this if it causes incompatibility with lambdas since that's what the documentation currently suggests to use. And I don't think `options` should be a mandatory argument either since most users will not need it.

Thanks for the PR!

vsevolod:
You're right. I've added arity condition to keep backward compatibility for lambdas.
Also, an additional `context` option has been added to the `counter_culture_fix_counts` options. So, it's optional - should we pass it or not.

magnusvk (approved):
Great—thanks for the update!

magnusvk:
Released as gem version 3.5.0

magnusvk:
Thanks for the contribution!

PR #376, opened 2023-08-26
Remove condition when Rails.version < 4.0.0 in user.rb

ken3ypa:
Removed unnecessary conditional branching from user.rb.
Rails.version < "4.0.0" was dropped in https://github.com/magnusvk/counter_culture/pull/252


magnusvk:
Thanks for the clean-up!

PR #380, opened 2023-10-12
README: add note for start/finish options for counter_culture_fix_counts

FunkyloverOne:
I was lucky to spot this in my specs. :sweat_smile:


magnusvk:
I didn't realize that's how this works, but looking at the code it appears you are right. That's pretty confusing! Thanks for the contribution.

PR #383, opened 2023-12-25
Fix touching for counted models without updated_at

dlackty:
We recently encountered a bug that prevents object from being touched correctly. 

When counter_culture tries to find touch column, it uses counted model for `timestamp_attributes_for_update_in_model`. However, counted model might have no updated_at column, and we should check model with counters to find correct timestamp attributes.

This issue is illustrated in my first commit, and fixed in second commit.


magnusvk:
Thank you! Released as gem version 3.5.1.

PR #386, opened 2024-01-10
Low level ActiveRecord attribute assignment on `previous_model`

### LINKED ISSUE #384: Call to `previous_model` is triggering `transaction_changed_attributes` callbacks causing changes to be overwritten
Hi there,

I'm reporting an issue here after observing potential complications with the `previous_model` call. It seems that this call has caused other problems before, and I wanted to bring it to your attention. Additionally, I'm reaching out to @dylanahsmith, the author of the [ArTransactionChanges](https://github.com/dylanahsmith/ar_transaction_changes/) gem, to gather his insights on this matter.

As you may know, the `counter_culture` gem utilizes a [combination of the `dup` call and direct assignment of previous changes](https://github.com/magnusvk/counter_culture/blob/74da0105539904a935f8251a03b9bb25ba99aac5/lib/counter_culture/counter.rb#L311) to track changes to the monitored attributes. These changes are later used to compute differences applied to the counter attributes in the parent model.

The issue here arises from duplicating an ActiveRecord model and then assigning new values to the tracked attributes. This triggers gems that intercept the [write_attribute](https://apidock.com/rails/ActiveRecord/AttributeMethods/write_attribute) call, such as ArTransactionChanges. Specifically, ArTransactionChanges detects a change but removes it from the set of changes recorded at the end. For example, if I set the quantity from 0 to 1, `transaction_changed_attributes` tracks that change. However, due to the subsequent `dup` & assign call, there is an action that negates the effects of the first one, resulting in the removal of that change from the set.

In conclusion, I'm questioning whether the strategy of `dup` and assign is the most suitable one. While it currently causes issues with the ArTransactionChanges gem, I believe there may be similar problems with other gems that intercept the `write_attribute` behavior.

Your attention to this matter is much appreciated.

Thank you.

jfloff:
Fixes #384


magnusvk:
Tests are green, so I guess this is good to go? I'm trying to think through whether there might be any unintended consequences of this. I can't really think of any, but did you go through that thought exercise? I just want to make sure I don't release this as a patch release and then it leads to issues for other folks.

jfloff:
I actually did that exercise while developing the fix, and the only thing I can think of is a situation where they were not aware of a bug similar to the one I reported, and now with the correct behavior in place something might change on their end. 
Otherwise, I don't think so, since the `previous_model` is such an internal `counter_culture` thing that people should not really build around IMO.

magnusvk:
Yeah, I agree. Fingers crossed we're right. 😄 I just released this as gem version 3.5.2.

Issue #374, opened 2023-08-15
STI and Polymorphic issue with fix counts method on self-referential models


attenzione:
Will simplify a lot
```ruby
class Scorable < ApplicationRecord
  self.table_name = 'players'
end

class Player < Scorable
  belongs_to :team

  counter_culture :team
end

class Team < Scorable
  has_many :players
end
```

CRUD operations are normally updating `players_count`.
But If execute `Player.counter_culture_fix_counts` all counters will be set to 0.

I've reviewed the code and found a mistake in the table name:
On lines [273](https://github.com/magnusvk/counter_culture/blob/master/lib/counter_culture/reconciler.rb#L273) and [279](https://github.com/magnusvk/counter_culture/blob/master/lib/counter_culture/reconciler.rb#L279) in `joins_sql` condition for STI and Polymorphic, table name is incorrect, because it must be `target_table_alias` instead of `target_table` as it seen in around lines (this is explicitly written on line [243](https://github.com/magnusvk/counter_culture/blob/master/lib/counter_culture/reconciler.rb#L243))

magnusvk:
Can you put together a PR with that fix and a test case that currently fails on master? Thanks for the report.

ramonpm:
This also affecting me, and due to no action in here, for a long time already, I'm taking it over and will implement the fix shortly.

attenzione:
Thanks @ramonpm. I have an email as task in my inbox since day one, but due to high demand on frontend work can't reach this.

ramonpm:
PR open, please review it :point_up:

PR #388, opened 2024-02-14
Bugfix 374 - Correct polymorphic table name alias reference in join clauses

ramonpm:
This fixes https://github.com/magnusvk/counter_culture/issues/374

A screencast explaining it all if you want to watch it https://www.loom.com/share/feb8cb3597be4d6aa9e314ecd532210c


magnusvk (approved):

magnusvk:
Thanks for the fix—released as gem version 3.5.3

Issue #389, opened 2024-02-14
Adding :if / :unless with predicate method symbol for conditional counter caches


tatethurston:
👋 Hey thanks for the great gem.

I'd like to propose supporting a conditional API similar to [ActiveRecord's callbacks](https://guides.rubyonrails.org/active_record_callbacks.html#using-if-and-unless-with-a-symbol).

Today, a [conditional counter cache](https://github.com/magnusvk/counter_culture?tab=readme-ov-file#conditional-counter-cache is implemented as follows:

```ruby
class Product < ActiveRecord::Base
  belongs_to :category
  counter_culture :category, column_name: proc {|model| model.special? ? 'special_count' : nil }
end

class Category < ActiveRecord::Base
  has_many :products
end
```

I would like to propose the following shorthand:

```ruby
class Product < ActiveRecord::Base
  belongs_to :category
  counter_culture :category, if: :special?
end

class Category < ActiveRecord::Base
  has_many :products
end
```

If you're amenable to that, I'd be happy to look into implementing the if/unless functionality, relevant specs, and interplay with the existing `column_name`. If not, no worries.

magnusvk:
That looks like a neat short-hand, yeah—I'd be happy to look at a PR for that. Thanks!

PR #390, opened 2024-02-17
Add conditional api

tatethurston:
Adds a conditional API similar to [ActiveRecord's Callbacks](https://guides.rubyonrails.org/active_record_callbacks.html#conditional-callbacks).

Resolves https://github.com/magnusvk/counter_culture/issues/389.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,24 @@ Now, the ```Category``` model will keep the counter cache in ```special_count```
 
 If you would like to use this with `counter_culture_fix_counts`, make sure to also provide [the `column_names` configuration](#handling-dynamic-column-names).
 
+### Conditional counter cache shorthand
+
+You may also use a conditional API identical to [ActiveRecord's Callbacks](https://guides.rubyonrails.org/active_record_callbacks.html#conditional-callbacks) to conditionally count columns:
+
+```ruby
+class Product < ActiveRecord::Base
+  belongs_to :category
+  counter_culture :category, column_name: :special_count, if: :special?
+end
+
+class Category < ActiveRecord::Base
+  has_many :products
+end
+
+Now, the ```Category``` model will keep the counter cache in ```special_count``` up-to-date. Only products where ```special?``` returns true will affect the special_count.
+
+If you would like to use this with `counter_culture_fix_counts`, make sure to also provide [the `column_names` configuration](#handling-dynamic-column-names).
+
 ### Temporarily skipping counter cache updates
 
 If you would like to temporarily pause counter_culture, for example in a backfill script, you can do so as follows:
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -17,6 +17,8 @@ module CounterCulture
       @delta_magnitude = options[:delta_magnitude] || 1
       @with_papertrail = options.fetch(:with_papertrail, false)
       @execute_after_commit = options.fetch(:execute_after_commit, false)
+      @if_conditions = Array.wrap(options[:if])
+      @unless_conditions = Array.wrap(options[:unless])
 
       if @execute_after_commit
         begin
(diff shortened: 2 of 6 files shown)


tatethurston:
Here's a first pass at #389. LMK what tweaks you'd like

lib/counter_culture/counter.rb
```diff
@@ -49,78 +51,79 @@ def change_counter_cache(obj, options)
...
-
-        klass = relation_klass(relation, source: obj, was: options[:was])
-
-        # MySQL throws an ambiguous column error if any joins are present and we don't include the
-        # table name. We isolate this change to MySQL because sqlite has the opposite behavior and
-        # throws an exception if the table name is present after UPDATE.
-        quoted_column = if klass.connection.adapter_name == 'Mysql2'
-                          "#{klass.quoted_table_name}.#{model.connection.quote_column_name(change_counter_column)}"
+      return unless conditions_allow_change?(obj)
+      return unless id_to_change && change_counter_column
```
tatethurston, line 55: lots of noise here but I the only change is switching to a guard clause since there are two now

README.md
```diff
@@ -190,6 +190,24 @@ Now, the ```Category``` model will keep the counter cache in ```special_count```
...
 If you would like to use this with `counter_culture_fix_counts`, make sure to also provide [the `column_names` configuration](#handling-dynamic-column-names).
 
+### Conditional counter cache shorthand
+
+You may also use a conditional API identical to [ActiveRecord's Callbacks](https://guides.rubyonrails.org/active_record_callbacks.html#conditional-callbacks) to conditionally count columns:
+
+```ruby
+class Product < ActiveRecord::Base
+  belongs_to :category
+  counter_culture :category, column_name: :special_count, if: :special?
```
magnusvk, line 200: I think we should have an example with `unless` as well.

lib/counter_culture/counter.rb
```diff
@@ -17,6 +17,8 @@ def initialize(model, relation, options)
       @delta_magnitude = options[:delta_magnitude] || 1
       @with_papertrail = options.fetch(:with_papertrail, false)
       @execute_after_commit = options.fetch(:execute_after_commit, false)
+      @if_conditions = Array.wrap(options[:if])
+      @unless_conditions = Array.wrap(options[:unless])
```
magnusvk, line 21: Here's a different idea, I'd love to discuss whether that's a better approach: What if we took what's passed into `options[:if]` or `options[:unless]` and "compile" that into a `Proc` to pass to use for `column_name`? This is pseudo-code, but something like
```
if options[:if].present?
  @counter_cache_name = proc  do |model|
    Array.wrap(options[:if]).inject(:&&) { |method| model.public_send(method) }
  end
end
```

The advantage I see of that is that it simplifies the code paths—there's still only one code path to consider and `if` and `unless` truly just become shorthand.

Is that better? Or does that make the code more complicated?

lib/counter_culture/counter.rb
```diff
@@ -342,5 +345,24 @@ def attribute_was(obj, attr)
...
+    def conditions_allow_change?(obj)
+      @if_conditions.all?      { |condition| evaluate_condition(condition, obj) } &&
+      @unless_conditions.none? { |condition| evaluate_condition(condition, obj) }
+    end
+
+    # Evaluate the conditions in the context of the object
+    def evaluate_condition(condition, obj)
+      case condition
+      when Symbol
+        obj.send(condition)
```
magnusvk, line 360:
```suggestion
        obj.public_send(condition)
```

spec/models/conditional_dependent_shorthand.rb
```diff
@@ -0,0 +1,8 @@
+class ConditionalDependentShorthand < ActiveRecord::Base
+  belongs_to :conditional_main
+  scope :condition, -> { where(condition: true) } 
+
+  counter_culture :conditional_main, if: :condition?, column_names: -> { {
```
magnusvk, line 5: We'd need a test for `unless` and for the case of passing a `Proc` as well.

spec/counter_culture_spec.rb
```diff
@@ -1711,6 +1712,24 @@ def yaml_load(yaml)
     ConditionalMain.find_each { |main| expect(main.conditional_dependents_count).to eq(main.id % 2 == 0 ? 3 : 0) }
   end
 
+  it "should correctly fix the counter caches for thousands of records when counter is conditional using :if/:unless" do
```
magnusvk, line 1715: We'd need a test for the normal updating on insert / update / delete as well, right? I don't even think we change `fix_counts` in this PR, that's just based on `column_names` so I don't even think this test here adds any test coverage we don't already have.

README.md
tatethurston: Agreed, just wanted to get alignment around the general approach before exploading out the variations

lib/counter_culture/counter.rb
tatethurston: Yeah that was the approach I considered originally, but then we can't specify a custom column name when using `if`/ `unless`.

spec/models/conditional_dependent_shorthand.rb
tatethurston: Yep, and when passing an array of symbols for if / unless, and when using if and unless together.

spec/counter_culture_spec.rb
tatethurston: Yeah I agree with sidestepping the fix counts logic since that's orthogonal. I tried to think of a way to have fix_count work well out of the box with this approach, but it would require counting in ruby instead of sql.

lib/counter_culture/counter.rb
magnusvk: Can you say a bit more about that? I don't think I follow why that doesn't work.

tatethurston: Sure, if I’m understanding you correctly that we “compile” into a proc passed internally to column_name, then we can’t do the following:

column_name: :foo_count, if: :special?

Because our compiled proc for if would overwrite the :foo_count symbol

magnusvk: Couldn't we use the value passed as the name in the proc we compile the `if` / `unless` into?

tatethurston: Yeah, you’re right. Nice, I’ll take a look at refactoring to that.

README.md
tatethurston: Specifically, symbol, proc, array of symbols, array of procs

lib/counter_culture/counter.rb
tatethurston: Updated, LMK what you think. Then I'll look at updating the tests

README.md
magnusvk: Yeah, I think so

lib/counter_culture/counter.rb
```diff
@@ -9,7 +9,25 @@ def initialize(model, relation, options)
...
+            model.public_send(condition)
+          when Proc
+            model.instance_exec(&condition)
+          else
+            raise ArgumentError, "Condition must be a symbol or a proc"
+          end
+        end
+
+        conditions_allow_change? = Array.wrap(options[:if]).all?(&conditions) &&
+          Array.wrap(options[:unless]).none?(&conditions)
```
magnusvk, line 25: Looks pretty neat!

lightyrs:
Hi @magnusvk — is this pending merge? What is outstanding here?

tatethurston:
> Hi @magnusvk — is this pending merge? What is outstanding here?

I’ve let this slip — are you interested in this functionality? If so, happy to prioritize getting this mergeable

magnusvk:
Refreshing my memory on this—it looks like there's a number of outstanding comment threads. The biggest issue is probably that it sounds like this doesn't work with `counter_culture_fix_counts`? What's the intent there?

tatthurs:
> Refreshing my memory on this—it looks like there's a number of outstanding comment threads. The biggest issue is probably that it sounds like this doesn't work with `counter_culture_fix_counts`? What's the intent there?

Yeah https://github.com/magnusvk/counter_culture/pull/390#discussion_r1496740016

tatethurston:
> Refreshing my memory on this—it looks like there's a number of outstanding comment threads. The biggest issue is probably that it sounds like this doesn't work with `counter_culture_fix_counts`? What's the intent there?

Previous discussion: https://github.com/magnusvk/counter_culture/pull/390#discussion_r1496740016.

I need to refresh my memory here, but I don't immediately see a reason why we couldn't make this work similar to the way something like `column_name: proc {|model| model.special? ? 'special_count' : nil } ` works today. Identical to that case, the user burden would be determining how to maximize performance for the fix_counts functionality.

Issue #354, opened 2022-06-05
Multiple counters, single query


thimmaiah:
Is it possible to generate a single query and update the counters ? I have many counters and the number of queries is piling up.

Thanks in advance

Singh212dab:
yes but not possible to mix counter query and non counter query in cassandra.

thimmaiah:
How to accomplish this, for mysql like DBs? Currently I have 10 counters which generate 10 different queries.

Singh212dab:
sorry mate no idea about it ,you can go for Google . only thing I know that it is possible

magnusvk:
This isn't currently possible. If you wanted to open a PR that adds this functionality I'd be happy to look at it.

marinfr:
https://github.com/magnusvk/counter_culture/pull/393

Issue #307, opened 2021-02-04
Ability to batch multiple counter cache updates for same row


lightyrs:
👋 @magnusvk

I've been using this great gem for awhile at joinsubtext.com, however, have recently been running into deadlocking issues as we scale. This idea is a little different than the other open deadlocking issue #263.

I have multiple counter caches defined on a model like this:

```rb
class Message < ApplicationRecord
  belongs_to :broadcast

  counter_culture :broadcast, column_name: proc { |model| model.draft? ? nil : "messages_count" }
  counter_culture :broadcast, column_name: proc { |model| model.delivered? ? "delivered_messages_count" : nil }
end
```

So when I look at the postgres logs and see something like below, I wonder if it would be possible to combine these updates into a single `UPDATE`.

```log
Process 26355: UPDATE "broadcasts" SET "messages_count" = COALESCE("messages_count", 0) + 1 WHERE "broadcasts"."id" = $1

Process 29749: UPDATE "broadcasts" SET "delivered_messages_count" = COALESCE("delivered_messages_count", 0) + 1 WHERE "broadcasts"."id" = $1
```

Thanks for your time.

magnusvk:
👋🏻—small world! 

This is a great idea! These can and should be consolidated. However taking a quick peek here it looks like that's a pretty big refactor from the current code. I'd be open to a pull request if you have time to work on this! :)

lightyrs:
@magnusvk Thanks for the quick reply. I will look into an implementation and submit a PR if I come up with something decent.

magnusvk:
Closing for inactivity.

marinfr:
https://github.com/magnusvk/counter_culture/pull/393

PR #393, opened 2024-06-05
Combine multiple updates into a single SQL per target row (N+1)

marinfr:
### Description

Title says it all 😄 

This had been brewing in my mind for quite a while and I finally got to implement it. I wanted it to be completely opt-in and not affect the normal flow. Implementation is not ideal but I felt like I didn't have too many options based on the current level of coupling and abstractions. I think it's abstracted enough to make it easy to maintain and hard to break even in the current state. Am open for suggestions though!

**Fixes:**
https://github.com/magnusvk/counter_culture/issues/307
https://github.com/magnusvk/counter_culture/issues/354

### TL;DR

<details><summary>Sums multiple counter updates for single column of a single target record</summary>

```sql
-- Before
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1 WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1 WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1 WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) - 1 WHERE `authors`.`id` = 1

-- After
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 2 WHERE `authors`.`id` = 1
```

```sql
-- Before
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) - 1 WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) - 1 WHERE `authors`.`id` = 1

-- After
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + -2 WHERE `authors`.`id` = 1
```

```sql
-- Before
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1 WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) - 1 WHERE `authors`.`id` = 1

-- After
No query
```

</details>

<details><summary>Combines multiple updates for multiple columns of a single target record</summary>

```sql
-- Before
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1, updated_at = '2024-06-06 01:00:00' WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1, updated_at = '2024-06-06 02:00:00' WHERE `authors`.`id` = 1

-- After
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 2, updated_at = '2024-06-06 02:00:00' WHERE `authors`.`id` = 1
```

```sql
-- Before
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1 WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`works_count` = COALESCE(`authors`.`works_count`, 0) + 1 WHERE `authors`.`id` = 1

-- After
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1, `authors`.`works_count` = COALESCE(`authors`.`works_count`, 0) + 1 WHERE `authors`.`id` = 1
```

</details>

</details>

<details><summary>Combines multiple updates of multiple target records</summary>

```sql
-- Before
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) - 1 WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) - 1 WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) - 1 WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1 WHERE `authors`.`id` = 2
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1 WHERE `authors`.`id` = 2
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 1 WHERE `authors`.`id` = 2

-- After
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + -3 WHERE `authors`.`id` = 1
UPDATE `authors` SET `authors`.`books_count` = COALESCE(`authors`.`books_count`, 0) + 3 WHERE `authors`.`id` = 2
```

</details>

<details><summary>Executes aggregated updates before commit</summary>

```ruby
ActiveRecord::Base.transaction do
  CounterCulture.aggregate_counter_updates do
    # list of updates
  end
end
```

</details>

<details><summary>Executes aggregated updates after commit</summary>

```ruby
CounterCulture.aggregate_counter_updates do
  ActiveRecord::Base.transaction do
    # list of updates
  end
end
```

</details>

### Implementation

I needed a higher domain of visibility for this mechanism to work, that's why I added the `CounterCulture.aggregate_counter_updates` method that receives a block with DB updates happening within it. Users passing code as blocks to this method means they decide where the updates start and where they end and whether the aggregated SQL queries should be executed before or after commit.

```ruby
CounterCulture.aggregate_counter_updates do
  ActiveRecord::Base.transaction do
    # list of updates
  end
end # => executes aggregated SQLs after COMMIT
```

The method initialises some `Thread.current` variables:
`aggregate_counter_updates` - boolean, used to change how the `after_*` AR callbacks behave (store update instead of executing the SQL immediately)
`aggregated_updates` - object, stores the updates per table and target record
`primary_key_map` - object, stores the primary keys for every table

After the variables are initialised, the user-defined block is executed, which hopefully added some updates to `aggregated_updates`. It iterates through all updates and groups them into single SQLs per target record. The SQLs get executed.

After that, the variables are reset and the response of the user-defined block is returned.

If `Thread.current[:aggregate_counter_updates]` is `true`, `change_counter_cache` will no longer execute SQLs immediately (**except for Papertrail callbacks**), but rather:
- for counter cache columns - stores the value of the first update as a temporary counter in a `table -> record ID -> column to be updated/operation` key, which then gets incremented or decremented by subsequent updates of the same column of the same record in the same table. The final value of this temporary counter will be used in the final SQL query.
- for timestamps - stores the value of the first update as a temporary value in a `table -> record ID -> column to be updated/operation` key, which then gets overwritten by subsequent updates of the same column of the same record in the same table. Because there's a delay between the value added to the cache and the SQL being executed, I couldn't pass the timestamp as a string. I opted to pass the timestamp-generating logic as a block, so I could have the freshest timestamp for the SQL request.

I wrote the basic tests to see that everything works and cherry-picked some more complex scenarios to see if they pass with the aggregated logic. I think anything else would be a bit overkill, but let me know if I missed some tests 🤷‍♂️

### DIFF
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -15,6 +15,35 @@ module CounterCulture
     yield(self) if block_given?
     self
   end
+
+  def self.aggregate_counter_updates
+    return unless block_given?
+
+    Thread.current[:aggregate_counter_updates] = true
+    Thread.current[:aggregated_updates] = {}
+    Thread.current[:primary_key_map] = {}
+
+    yield
+
+    # aggregate the updates for each target and execute SQL queries
+    Thread.current[:aggregated_updates].each do |klass, attrs|
+      attrs.each do |rec_id, updates|
+        updated_columns = updates.map do |operation, value|
+          %Q{#{operation} #{value.is_a?(Proc) ? "'#{value.call}'" : value}} unless value == 0
+        end.compact
+
+        klass
+          .where(Thread.current[:primary_key_map][klass] => rec_id)
+          .update_all(updated_columns.join(', '))
+      end
+    end
+
+    nil
+  ensure
+    Thread.current[:aggregate_counter_updates] = false
+    Thread.current[:aggregated_updates] = nil
+    Thread.current[:primary_key_map] = nil
+  end
 end
 
 # extend ActiveRecord with our own code here
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -28,15 +28,15 @@ module CounterCulture
     # aggregate the updates for each target record and execute SQL queries
     Thread.current[:aggregated_updates].each do |klass, attrs|
       attrs.each do |rec_id, updates|
-        updated_columns = updates.map do |operation, value|
+        update_snippets = updates.map do |operation, value|
           value = value.call if value.is_a?(Proc)
           %Q{#{operation} #{value.is_a?(String) ? "'#{value}'" : value}} unless value == 0
         end.compact
 
-        if updated_columns.any?
+        if update_snippets.any?
           klass
             .where(Thread.current[:primary_key_map][klass] => rec_id)
-            .update_all(updated_columns.join(', '))
+            .update_all(update_snippets.join(', '))
         end
       end
     end
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -25,20 +25,21 @@ module CounterCulture
 
     yield
 
-    # aggregate the updates for each target and execute SQL queries
+    # aggregate the updates for each target record and execute SQL queries
     Thread.current[:aggregated_updates].each do |klass, attrs|
       attrs.each do |rec_id, updates|
         updated_columns = updates.map do |operation, value|
-          %Q{#{operation} #{value.is_a?(Proc) ? "'#{value.call}'" : value}} unless value == 0
+          value = value.call if value.is_a?(Proc)
+          %Q{#{operation} #{value.is_a?(String) ? "'#{value}'" : value}} unless value == 0
         end.compact
 
-        klass
-          .where(Thread.current[:primary_key_map][klass] => rec_id)
-          .update_all(updated_columns.join(', '))
+        if updated_columns.any?
+          klass
+            .where(Thread.current[:primary_key_map][klass] => rec_id)
+            .update_all(updated_columns.join(', '))
+        end
       end
     end
-
-    nil
   ensure
     Thread.current[:aggregate_counter_updates] = false
     Thread.current[:aggregated_updates] = nil
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -23,7 +23,7 @@ module CounterCulture
     Thread.current[:aggregated_updates] = {}
     Thread.current[:primary_key_map] = {}
 
-    yield
+    result = yield
 
     # aggregate the updates for each target record and execute SQL queries
     Thread.current[:aggregated_updates].each do |klass, attrs|
@@ -40,6 +40,8 @@ module CounterCulture
         end
       end
     end
+
+    result
   ensure
     Thread.current[:aggregate_counter_updates] = false
     Thread.current[:aggregated_updates] = nil
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -18,6 +18,9 @@ module CounterCulture
       @with_papertrail = options.fetch(:with_papertrail, false)
       @execute_after_commit = options.fetch(:execute_after_commit, false)
 
+      # we don't use Rails' update_counters because we support changing the timestamp
+      @updates_to_execute = []
+
       if @execute_after_commit
         begin
           require 'after_commit_action'
@@ -71,13 +74,11 @@ module CounterCulture
 
         column_type = klass.type_for_attribute(change_counter_column).type
 
-        # we don't use Rails' update_counters because we support changing the timestamp
-        updates = []
         # this updates the actual counter
         if column_type == :money
-          updates << "#{quoted_column} = COALESCE(CAST(#{quoted_column} as NUMERIC), 0) #{operator} #{delta_magnitude}"
+          assemble_money_counter_update(klass, id_to_change, quoted_column, operator, delta_magnitude)
         else
-          updates << "#{quoted_column} = COALESCE(#{quoted_column}, 0) #{operator} #{delta_magnitude}"
+          assemble_counter_update(klass, id_to_change, quoted_column, operator, delta_magnitude)
         end
         # and here we update the timestamp, if so desired
         if touch
@@ -89,11 +90,20 @@ module CounterCulture
             timestamp_columns << touch
           end
           timestamp_columns.each do |timestamp_column|
-            updates << "#{timestamp_column} = '#{current_time.to_formatted_s(:db)}'"
+            assemble_timestamp_update(
+              klass,
+              id_to_change,
+              timestamp_column,
+              -> { klass.send(:current_time_from_proper_timezone).to_formatted_s(:db) }
+            )
           end
         end
 
         primary_key = relation_primary_key(relation, source: obj, was: options[:was])
+
+        if Thread.current[:aggregate_counter_updates]
+          Thread.current[:primary_key_map][klass] ||= primary_key
+        end
 
         if @with_papertrail
           instance = klass.where(primary_key => id_to_change).first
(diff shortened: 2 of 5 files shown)


marinfr:
@magnusvk I'm not sure why some SQLite tests fail, but doesn't look like it's relevant to the PR?

lightyrs:
@magnusvk What do you think of this? Seems like a good starting point.

magnusvk:
@marinfr @marinmeister can you rebase on the latest master to pull in the sqlite fix?

marinfr:
@magnusvk All is good now 👍

lib/counter_culture/counter.rb
```diff
@@ -344,5 +358,60 @@ def attribute_was(obj, attr)
...
+          "#{update} +",
+          operator == '+' ? delta_magnitude : -delta_magnitude
+        )
+      else
+        "#{update} #{operator} #{delta_magnitude}"
+      end
+    end
+
+    def assemble_counter_update(klass, id_to_change, quoted_column, operator, delta_magnitude)
+      update = "#{quoted_column} = COALESCE(#{quoted_column}, 0)"
```
magnusvk, line 378: I think `assemble_money_counter_update` and `assemble_counter_update` are the same except for the first line? Can we pull the common code out into a third method that both of these call?

spec/counter_culture_spec.rb
```diff
@@ -2967,4 +2967,267 @@ def mess_up_counts
...
+
+      expect(user.reviews_count).to eq(1)
+      expect(user.using_count).to eq(1)
+      expect(user.review_approvals_count).to eq(5)
+      expect(user.dynamic_delta_count).to eq(1)
+      expect(user.custom_delta_count).to eq(3)
+      expect(product1.reviews_count).to eq(1)
+      expect(product1.rexiews_count).to eq(1)
+
+      CounterCulture.aggregate_counter_updates do
```
magnusvk, line 2999: How do you feel about using that new `expect_queries` helper in some of these tests?

spec/counter_culture_spec.rb
```diff
@@ -2967,4 +2967,267 @@ def mess_up_counts
...
+      expect(po.total_amount).to eq(100.0)
+
+      CounterCulture.aggregate_counter_updates do
+        po.purchase_order_items.destroy_all
+      end
+
+      po.reload
+      expect(po.total_amount).to eq(0.0)
+    end
+  end
```
magnusvk, line 3232: I think it might be nice to have one "integration-style" test where we do lots of things and make sure that the results end up correct. Maybe some combination of:
* Multiple changes that update the same column, same id
* Other multiple changes that update a different column on the same id, and make that one a `delta_column` rather than a plain counter
* Then some changes on one of the above columns, but for a different id
* A money column change
* A timestamp column change

I have no indication that this would surface anything that won't work, but for aggregation logic like this it's nice to see it all pulled together. And if you combine it with `expect_queries` then we can even make sure that it's aggregating as expected? That last part feels optional, though.

lib/counter_culture.rb
```diff
@@ -15,6 +15,38 @@ def self.config
...
+
+    result = yield
+
+    # aggregate the updates for each target record and execute SQL queries
+    Thread.current[:aggregated_updates].each do |klass, attrs|
+      attrs.each do |rec_id, updates|
+        update_snippets = updates.map do |operation, value|
+          value = value.call if value.is_a?(Proc)
+          %Q{#{operation} #{value.is_a?(String) ? "'#{value}'" : value}} unless value == 0
+        end.compact
```
benjaminwols, line 34: @magnusvk Is there a reason this gem still supports Ruby 2.6 (which is EOL since 31 Mar 2022)?
Ruby 2.7 added filter_map, so if the gem will drop support for 2.6, this could be simplified.
```suggestion
        update_snippets = updates.filter_map do |operation, value|
          value = value.call if value.is_a?(Proc)
          %Q{#{operation} #{value.is_a?(String) ? "'#{value}'" : value}} unless value == 0
        end
```

lib/counter_culture/counter.rb
marinmeister: Done

spec/counter_culture_spec.rb
marinmeister: I wrote a bigger test for this

spec/counter_culture_spec.rb
marinmeister: I added two more complex tests:
1. to test that counter increments by 0 are ignored
2. a test case that tests:
2.1 Multiple changes that update the same column, same id (both on user and product)
2.2 Other multiple changes that update a different column on the same id, and make that one a delta_column rather than a plain counter (user)
2.3 A timestamp column change (product)
2.4 Then some changes on one of the above columns, but for a different id (you have 2 users and 2 products)

I left the money counter test be on its own (I added a test for it previously)

lib/counter_culture.rb
magnusvk: No reason, really, I tend to drop support when supporting the old version starts creating headaches, rather than when a version goes EOL. Aka I would be very open to dropping support for 2.6. That being said, I'm not sure I prefer `filter_map` over `map.compact`—I find the former to be just as readable, if not more so?

spec/counter_culture_spec.rb
magnusvk: Yeah, this looks great.

magnusvk:
Fantastic, thanks for the updates. I think the only thing this is missing is documentation in the readme, then it's good to merge and release. Can you add something there?

marinfr:
> Fantastic, thanks for the updates. I think the only thing this is missing is documentation in the readme, then it's good to merge and release. Can you add something there?

Done:
image

lib/counter_culture.rb
benjaminwols: I have been using filter_map for a while now, so for me it feels naturally to use it. I dislike chaining methods to end, but that is just a personal preference. 
I was curious if there was a reason to keep supporting older Ruby versions, but it is totally fine for me to keep it as is and decide on dropping 2.6 at a later stage when it brings more benefit.

magnusvk (approved):

magnusvk:
Amazing, thank you so much for the contribution! I just released this as gem version 3.7.0.

PR #398, opened 2024-10-02
feat: use `with_connection` where possible

flood4life:
https://github.com/rails/rails/pull/51349 - this allows rails 7.2 applications to raise an error or emit a deprecation warning whenever `ActiveRecord::Base.connection` is used.

The PR makes sure that in Rails 7.1 and after we use the new, preferred method of obtaining a connection.


magnusvk:
Thanks for the contribution! I just added testing against Rails 7.2 to the latest master, can you merge master or rebase to pull that in before we merge this?

lib/counter_culture/with_connection.rb
```diff
@@ -0,0 +1,19 @@
...
+    attr_reader :recipient
+
+    def call
+      if recipient.respond_to?(:with_connection)
+        recipient.with_connection do |connection|
+          yield connection
+        end
+      else
+        yield recipient.connection
+      end
```
magnusvk, line 16:
```suggestion
      if Gem::Version.new(Rails.version) < Gem::Version.new('7.2.0')
        yield recipient.connection
      else
        recipient.with_connection do |connection|
          yield connection
        end
      end
```

How do you feel about using the Rails version here instead to determine which method to call? In my head, if we're in Rails 7.2 or later, where `with_connection` should be available, it would be an error if `respond_to?(:with_connection)` was false. I realize I'm nitpicking and this probably doesn't matter in reality, but I _think_ I have a preference for using `Rails.version`. Do let me know if I'm missing something there though!

flood4life: I don't have a strong preference, so happy to switch to a Rails version check. One tiny change though, is `with_connection` is available from 7.1; 7.2 just adds the option to deprecate/disable the old API. So I would go with checking against 7.1 if you don't have any concerns? As a side note, we're already running this in a fairly big 7.1 production application, and haven't noticed any problems so far.

magnusvk: Yeah, let's switch it then. If it was added in 7.1, great, let's use that! The specs would let us know if it wasn't available. 😄

flood4life: The specs did catch something: apparently, in 7.1 `with_connection` is not available on `ActiveRecord::Base` class and its instance, but it available on `ConnectionPool`, so I've added a bit more logic to handle that.

magnusvk (approved):

magnusvk:
Thanks again! Released as gem version 3.8.0

dshorthouse:
I use ActiveRecord in a Sinatra environment and so `Rails.version` is not available. How might I continue making use of counter_culture?

dshorthouse:
That's awesome! Thanks for the quick turn-around.

PR #402, opened 2024-10-21
Use `ActiveRecord.version` instead of `Rails.version`

magnusvk:
Because it turns out that you can use ActiveRecord outside of Rails (today I learned!)


magnusvk:
cc @dshorthouse

magnusvk:
Released as 3.8.2

dshorthouse:
Works great! Much appreciated @magnusvk.

Issue #403, opened 2024-12-23
Feature Request: Add option to use read replicas by default for count operations


ashishra0:
## Problem
Currently, when using counter_culture to fix counts, all read operations (including counting records) happen on the primary database. This can cause significant lock contentions and performance issues, especially when running counter_culture_fix_counts in background jobs with large datasets.

## Current Workaround
We currently have to explicitly specify using read replicas for each counter_culture_fix_counts call:

```ruby
Model.counter_culture_fix_counts(
  db_connection_builder: proc { |reading, block|
    if reading
      ApplicationRecord.connected_to(role: :reading, &block)
    else
      ApplicationRecord.connected_to(role: :writing, &block)
    end
  }
)
```

I was thinking something along these lines

```ruby
# In initializer
CounterCulture.configure do |config|
  config.use_read_replica = true
end

# Or as a default option in counter_culture itself
```

Apologies for asking if this feature already exists

magnusvk:
Yeah, that does seem like a simpler option, and I don't think it exists yet. 

I won't be able to work on a change like that myself, but would be happy to look at a PR implementing this.

ashishra0:
Okay great, I will try and spin up a PR for this! Thanks :)

PR #405, opened 2025-03-14
Add global configuration for read replica support

ashishra0:
## Description
Adds a global configuration option to use read replicas for counting operations in counter cache reconciliation. This simplifies the process of using read replicas - instead of configuring a custom connection builder for each call, users can now enable it globally.

For feature #403 

## Changes
- Added `use_read_replica` configuration option
- When enabled, automatically uses:
  - Read replica for counting operations
  - Primary database for updates
- Maintains compatibility with existing `db_connection_builder` option (custom builders take precedence)

## Usage
Enable read replica support globally:
```ruby
# config/initializers/counter_culture.rb
CounterCulture.configure do |config|
  config.use_read_replica = true
end
```

Previous approach (still supported):
```ruby
# Manual configuration per call
Product.counter_culture_fix_counts(
  db_connection_builder: proc{|reading, block|
    if reading
      Product.connected_to(role: :reading, &block)
    else
      Product.connected_to(role: :writing, &block)
    end
  }
)
```


ashishra0:
@magnusvk Does this make sense?

lib/counter_culture/with_connection.rb
```diff
@@ -6,22 +6,27 @@ def initialize(recipient)
 
     attr_reader :recipient
 
-    def call
+    def call(reading: false)
+      use_read_replica = CounterCulture.configuration.use_read_replica && reading
+      role = use_read_replica ? :reading : :writing
+
+      ActiveRecord::Base.connected_to(role: role) do
```
magnusvk, line 13: I think to get the tests to pass for older Rails versions we'll have to do this only if we're at Rails 7.1 or greater, right?

magnusvk:
I think this looks good, yeah—I'd be happy to add this. Could you make sure the tests pass for older Rails versions, too?

ashishra0:
> I think this looks good, yeah—I'd be happy to add this. Could you make sure the tests pass for older Rails versions, too?

Yeah I will get that sorted!

lib/counter_culture/with_connection.rb
ashishra0: Yeah that's right!

lib/counter_culture/configuration.rb
```diff
@@ -0,0 +1,21 @@
...
+  def self.configure
+    yield(configuration)
+  end
+
+  def self.reset_configuration
+    @configuration = Configuration.new
+  end
+
+  class Configuration
+    attr_accessor :use_read_replica
```
magnusvk, line 15: It would be great to have this raise an exception if someone tries to set it on an older Rails version. Is that simple to add? Not a requirement if not.

README.md
```diff
@@ -618,6 +618,39 @@ Image.counter_culture_fix_counts(polymorphic_classes: Product)
 Image.counter_culture_fix_counts(polymorphic_classes: [Product, Employee])
 ```
 
+## Using Read Replicas
+
+When using `counter_culture_fix_counts`, you can configure Counter Culture to use read replicas for counting operations, which can help reduce load on your primary database. Write operations will still use the primary database.
```
magnusvk, line 623:
```suggestion
When using `counter_culture_fix_counts`, you can configure counter_culture to use read replicas for counting operations, which can help reduce load on your primary database. Write operations will still use the primary database.
```

magnusvk (approved):

lib/counter_culture/configuration.rb
ashishra0: Yeah should be simple enough!

ashishra0: Pushed a commit for the same

lib/counter_culture/configuration.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+      @use_read_replica = value
+    end
+
+    private
+
+    def rails_supports_read_replica?
+      Rails::VERSION::STRING.to_f >= 6.1
+    rescue NameError
+      false
```
magnusvk, line 33:
```suggestion
      Gem::Requirement.new('>= 7.1.0').satisfied_by?(ActiveRecord.version)
```
I think it's 7.1, not 6.1 that's required? And if we use this same approach of checking the ActiveRecord version then I think that avoids needing to rescue `NameError`

ashishra0: Oh yes you're right!

magnusvk:
Released as gem version 3.9.0. Thanks for the contribution!

ashishra0:
Thank you so much for the quick review and merging it :)

Issue #236, opened 2018-10-29
Create and destroy do not reload associated object


cintamani:
## The problem 
When creating or destroying an object with a `counter_cache` association, the associated object will update the count correctly but its instance is not reloaded in the original object. 

Given 2 models: 
```
class Post < ApplicationRecord
  has_many :comments
end

class Comment < ApplicationRecord
  belongs_to :post 
  counter_culture :post 
end
```

I expect that: 
```
post = Post.last 
comment = Comment.create(post: post, body: "A comment.") 

expect(comment.post.comments_count).to eq(1)
```
without the need to `comment.post.reload`. 

The important things to note here is that if we use the rails `counter_cache` feature this test will pass.  
The `counter_culture` gem instead does not reload the instance of the relationship object, it needs to be manually reloaded. 
This is causing problems in our API environment where the JSON response are created from the instances and the counters do not reflect the persisted data. 

Is this an actual `bug` or a feature? Is there any way to tell `counter_culture` to reload the relation object instances after a create and destroy?

magnusvk:
I'm not sure if I'd call this a bug, but it would definitely be an improvement for this to happen without the need for a reload. That being said, I won't have time to work on this. But I'd love a PR to this effect!

cintamani:
thanks for the clarification @magnusvk :) I'll write some code and open a PR for this

night91:
@cintamani Did you find a way to overcome this? Because the problem you exposed is complicated sometimes.

vitalinfo:
does it something what was requested?
https://github.com/magnusvk/counter_culture/compare/master...vitalinfo:counter_culture:master?expand=1
if yes, I can finalize my PR with tests and other stuff

PR #407, opened 2025-04-14
Create and destroy update associated object counter

vitalinfo:
FIX: https://github.com/magnusvk/counter_culture/issues/236


magnusvk (approved):
Thank you—this looks great!

magnusvk:
Released as gem version 3.10.0.

maddieholtzer:
I believe that this may have introduced a breaking change for those using `counter_culture` without an explicit association marked. Have noted this in an issue [here](https://github.com/magnusvk/counter_culture/issues/408).

PR #410, opened 2025-04-25
Showcase of what is breaking for us in 3.10.0

maddieholtzer:
I have found a test change that showcases the issue we are experiencing. This breaks `spec/counter_culture_spec.rb:2077` on 3.10.0 but that test then passes with this new code after commenting out the new `assign_to_associated_object` call added in [this pr](https://github.com/magnusvk/counter_culture/pull/407).

Not 100% sure yet why this change breaks this implementation (delegations on multi level counter caches)

### DIFF
--- a/spec/models/conversation.rb
+++ b/spec/models/conversation.rb
@@ -1,4 +1,8 @@
 class Conversation < ActiveRecord::Base
   belongs_to :candidate
+  delegate(
+    :candidate_profile,
+    to: :candidate
+  )
   counter_culture [:candidate, :candidate_profile]
 end


maddieholtzer:
[This circleci job](https://app.circleci.com/pipelines/github/magnusvk/counter_culture/173/workflows/5a82426f-788f-47b5-ba53-2f8656ac4477/jobs/9676) is one that shows the issue:

Screenshot 2025-04-25 at 14 47 56

spec/models/conversation.rb
```diff
@@ -1,4 +1,8 @@
 class Conversation < ActiveRecord::Base
   belongs_to :candidate
+  delegate(
+    :candidate_profile,
+    to: :candidate
+  )
```
magnusvk, line 6:
```suggestion
    has_one :candidate_profile, through: :candidate
```
Does it work if you do this instead?

maddieholtzer: So when trying this on our code base it fixes the issues specific to `counter_culture` but it then breaks a few other things for us so it appears we are actually relying on it being a delegation instead of a `has_one` although I haven't been able to pinpoint exactly why.

Issue #408, opened 2025-04-16
3.10.0 breaks our current implementation


maddieholtzer:
With the update to 3.10.0 we see that a substantial portion of our tests are breaking.

Doing some testing with this it appears that this may be because we are using `counter_culture` without explicitly adding the association.

When doing the upgrade with just:

```
#app/models/form_request.rb

counter_culture [:campaign]
```

we get this failure:

```
Failure/Error: let(:form_request) { create(:form_request) }
     
ActiveRecord::AssociationNotFoundError: Association named 'campaign' was not found on FormRequest; perhaps you misspelled it?
```

When updated to include the association:

```
belongs_to :campaign
counter_culture [:campaign]
```

things now work in `3.10.0`

Curious whether our current implementation was ever intended to be supported our not.

vitalinfo:
@maddieholtzer could you please provide more backtrace, because I can't reproduce such error 

Without `belongs_to` I get an exception:

```
  1) CounterCulture 
     Failure/Error: value = value.send(relation.shift)
     
     NoMethodError:
       undefined method `post' for an instance of PostLike
     # .../gems/ruby-3.3.1/gems/activemodel-6.0.6.1/lib/active_model/attribute_methods.rb:432:in `method_missing'
     # ./lib/counter_culture/counter.rb:202:in `foreign_key_value'
     # ./lib/counter_culture/counter.rb:50:in `change_counter_cache'
     # ./lib/counter_culture/extensions.rb:111:in `block in _update_counts_after_create'
     # ./lib/counter_culture/extensions.rb:109:in `each'
     # ./lib/counter_culture/extensions.rb:109:in `_update_counts_after_create'
     # ./lib/counter_culture/skip_updates.rb:8:in `_update_counts_after_create'
```

but this's definitely doesn't related to this changes https://github.com/magnusvk/counter_culture/pull/407 and as for me pretty expected

magnusvk:
@maddieholtzer I tried reproducing this, see #409. But you can see there that this fails both with and without the changes in #407.

My first instinct here is to say "I would have never expected that to work, and why don't you just add the association?" At the same time, if there's an easy fix for this I'm certainly not opposed to adding that.

maddieholtzer:
Hello @magnusvk 

I have just put up a pr showcasing the issue we are experiencing [here](https://github.com/magnusvk/counter_culture/pull/410)

The change in 3.10.0 appears to have broken implementations where delegations are used on multi-level counters.

PR #411, opened 2025-04-29
Fixed assign count to associated object when delegate

vitalinfo:
fix for https://github.com/magnusvk/counter_culture/issues/408

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -370,15 +370,26 @@ module CounterCulture
     def assign_to_associated_object(obj, relation, change_counter_column, operator, delta_magnitude)
       association_name = relation_reflect(relation).name
 
-      return unless obj.respond_to?(association_name)
+      association_object = association_object_for_assign(obj, association_name)
+      return if association_object.blank?
 
-      if obj.association(association_name).loaded? && (association_object = obj.public_send(association_name)).present?
-        association_object.assign_attributes(
-          change_counter_column =>
-            association_object_new_counter(association_object, change_counter_column, operator, delta_magnitude)
-        )
-        association_object_clear_change(association_object, change_counter_column)
+      association_object.assign_attributes(
+        change_counter_column =>
+          association_object_new_counter(association_object, change_counter_column, operator, delta_magnitude)
+      )
+      association_object_clear_change(association_object, change_counter_column)
+    end
+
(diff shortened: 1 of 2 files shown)


lib/counter_culture/counter.rb
```diff
@@ -370,15 +370,26 @@ def attribute_was(obj, attr)
...
+    end
+
+    def association_object_for_assign(obj, association_name)
+      unless obj.class
+            .reflect_on_all_associations
+            .detect { |assoc| assoc.name.to_sym == association_name.to_sym }
+            .present? &&
+         obj.association(association_name).loaded?
+        return
       end
```
magnusvk, line 390:
```suggestion
      if obj.class.reflect_on_all_associations.
          find { |assoc| assoc.name.to_sym == association_name.to_sym }.
          blank?
        # This means that this isn't defined as an association; either it
        # doesn't exist at all, or it uses delegate. In either case, we can't
        # update the count this way.
        return
      end
      if !obj.association(association_name).loaded?
        # If the association isn't loaded, no need to update the count
        return
      end
```

lib/counter_culture/counter.rb
```diff
@@ -381,11 +381,16 @@ def assign_to_associated_object(obj, relation, change_counter_column, operator,
...
-         obj.association(association_name).loaded?
+      if obj.class.reflect_on_all_associations.
+          find { |assoc| assoc.name.to_sym == association_name.to_sym }.
+          blank?
+        # This means that this isn't defined as an association; either it
+        # doesn't exist at all, or it uses delegate. In either case, we can't
+        # update the count this way.
+        return
+      end
+      if !obj.association(association_name).loaded?
```
vitalinfo, line 392: @magnusvk I think `unless` works here much better, but up to you

magnusvk: I always avoid `unless` unless it's just a single, simple condition—I find it much, much harder to understand with something like an `&&` in the condition because you have to invert everything there mentally.

magnusvk:
Thanks for the quick follow-up with the fix here! cc @maddieholtzer I just released this as gem version 3.10.1

maddieholtzer:
@magnusvk @vitalinfo This fixed things for us!!!!!! Thank you so so much I realize this is a pretty weird edge case.

PR #413, opened 2025-06-01
feat: allow composite primary keys for Rails v7.2+

### LINKED ISSUE #412: Unable to use counter_culture with composite primary keys
Hey there, trying to set up `counter_culture` with some many-to-many associations. 

```ruby
class Course
  has_many :students, through: :students_courses
  has_many :students_courses, primary_key: [:id, :curriculum_revision_id]
end

class Student
  has_many :courses, through: :students_courses
  has_many :students_courses
end

class StudentsCourse
  belongs_to :course, primary_key: [:id, :curriculum_revision_id]
  belongs_to :student

  counter_culture :student
  counter_culture :course
end
```

It seems that the lookup of the primary key goes fine:
```
(ruby) relation_primary_key(original_relation, source: obj, was: was)
["id", "curriculum_revision_id"]
```

but then since we got an array back, the [translation to a symbol fails](https://github.com/magnusvk/counter_culture/blob/v3.10.1/lib/counter_culture/counter.rb#L205):
```
(ruby) relation_primary_key(original_relation, source: obj, was: was).try(:to_sym)
nil
```
 and we get a `TypeError: nil is not a symbol nor a string`.

notchairmk:
Fixes #412 

This change allows `counter_culture` to be used for associations that make use of `primary_key/foreign_key` with array values.

I tested that this change doesn't affect backwards compatibility of existing functionality for older versions of activerecord, but the test setup is reliant on the format of composite primary keys supported in activerecord v7.2+.

Also this change is functional with and has tests for composite key functionality with:
- papertrail
- `aggregate_counter_updates`
- `counter_culture_fix_counts`

### DIFF
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -185,12 +185,7 @@ module CounterCulture
 
     # the string to pass to order() in order to sort by primary key
     def full_primary_key(klass)
-      primary_key = klass.primary_key
-      if primary_key.is_a?(Array)
-        primary_key.map { |pk| "#{klass.quoted_table_name}.#{pk}" }.join(', ')
-      else
-        "#{klass.quoted_table_name}.#{klass.quoted_primary_key}"
-      end
+      Array.wrap(primary_key).map { |pk| "#{klass.quoted_table_name}.#{pk}" }.join(', ')
     end
 
     # gets the value of the foreign key on the given relation
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -113,12 +113,8 @@ module CounterCulture
         end
 
         if @with_papertrail
-          instance = if primary_key.is_a?(Array)
-            conditions = composite_primary_key_conditions(primary_key, id_to_change)
-            klass.where(conditions).first
-          else
-            klass.where(primary_key => id_to_change).first
-          end
+          conditions = primary_key_conditions(primary_key, id_to_change)
+          instance = klass.where(conditions).first
 
           if instance
             if instance.paper_trail.respond_to?(:save_with_version)
@@ -143,13 +139,8 @@ module CounterCulture
 
         unless Thread.current[:aggregate_counter_updates]
           execute_now_or_after_commit(obj) do
-            if primary_key.is_a?(Array)
-              # handle composite primary keys
-              conditions = composite_primary_key_conditions(primary_key, id_to_change)
-              klass.where(conditions).update_all updates.join(', ')
-            else
-              klass.where(primary_key => id_to_change).update_all updates.join(', ')
-            end
+            conditions = primary_key_conditions(primary_key, id_to_change)
+            klass.where(conditions).update_all updates.join(', ')
 
             assign_to_associated_object(obj, relation, change_counter_column, operator, delta_magnitude)
           end
@@ -185,7 +176,7 @@ module CounterCulture
 
     # the string to pass to order() in order to sort by primary key
     def full_primary_key(klass)
-      Array.wrap(primary_key).map { |pk| "#{klass.quoted_table_name}.#{pk}" }.join(', ')
+      Array.wrap(klass.primary_key).map { |pk| "#{klass.quoted_table_name}.#{pk}" }.join(', ')
     end
 
     # gets the value of the foreign key on the given relation
@@ -205,14 +196,8 @@ module CounterCulture
         klass = relation_klass(first, source: obj, was: was)
         if foreign_key_value
           primary_key = relation_primary_key(first, source: obj, was: was)
-          if primary_key.is_a?(Array)
-            conditions = composite_primary_key_conditions(primary_key, foreign_key_value)
-            klass.where(conditions).first
-          else
-            klass.where(
-              "#{klass.table_name}.#{primary_key} = ?",
-              foreign_key_value).first
-          end
+          conditions = primary_key_conditions(primary_key, foreign_key_value)
+          klass.where(conditions).first
         end
       else
         obj
@@ -222,12 +207,7 @@ module CounterCulture
       end
 
       primary_key = relation_primary_key(original_relation, source: obj, was: was)
-      if primary_key.is_a?(Array)
-        # handle composite primary keys
-        return value ? primary_key.map { |key| value.public_send(key) } : nil
-      else
-        return value.try(primary_key.try(:to_sym))
-      end
+      Array.wrap(primary_key).filter_map { |pk| value.try(pk&.to_sym) }.presence
     end
 
     # gets the reflect object on the given relation
@@ -470,10 +450,10 @@ module CounterCulture
       end
     end
 
-    def composite_primary_key_conditions(primary_keys, fk_value)
-      primary_keys.each_with_index.map do |key, index|
-        [key, fk_value[index]]
-      end.to_h
+    def primary_key_conditions(primary_key, fk_value)
+      Array.wrap(primary_key)
+          .zip(Array.wrap(fk_value))
+          .to_h
     end
 
     def counter_update_snippet(update, klass, id_to_change, operator, delta_magnitude)
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -314,11 +314,6 @@ module CounterCulture
         return relation_klass(relation, source: source, was: was).try(:primary_key)
       end
 
-      # handle composite primary keys
-      if reflect.options[:primary_key].is_a?(Array)
-        return reflect.options[:primary_key]
-      end
-
       reflect.association_primary_key(klass)
     end
 
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -176,7 +176,7 @@ module CounterCulture
 
     # the string to pass to order() in order to sort by primary key
     def full_primary_key(klass)
-      Array.wrap(klass.primary_key).map { |pk| "#{klass.quoted_table_name}.#{pk}" }.join(', ')
+      Array.wrap(klass.quoted_primary_key).map { |pk| "#{klass.quoted_table_name}.#{pk}" }.join(', ')
     end
 
     # gets the value of the foreign key on the given relation
--- a/lib/counter_culture/counter.rb
+++ b/lib/counter_culture/counter.rb
@@ -207,7 +207,7 @@ module CounterCulture
       end
 
       primary_key = relation_primary_key(original_relation, source: obj, was: was)
-      Array.wrap(primary_key).filter_map { |pk| value.try(pk&.to_sym) }.presence
+      Array.wrap(primary_key).map { |pk| value.try(pk&.to_sym) }.compact.presence
     end
 
     # gets the reflect object on the given relation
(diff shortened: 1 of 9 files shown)


lib/counter_culture/reconciler.rb
```diff
@@ -271,8 +287,18 @@ def join_clauses(where)
               [target_table_key, source_table_key]
           end
 
-          joins_sql = "LEFT JOIN #{target_table} AS #{target_table_alias} "\
-            "ON #{source_table}.#{source_table_key} = #{target_table_alias}.#{target_table_key}"
+          if source_table_key.is_a?(Array) && target_table_key.is_a?(Array)
```
notchairmk, line 290: I'll clean this one up as well

lib/counter_culture/counter.rb
```diff
@@ -172,7 +176,7 @@ def counter_cache_name_for(obj)
 
     # the string to pass to order() in order to sort by primary key
     def full_primary_key(klass)
-      "#{klass.quoted_table_name}.#{klass.quoted_primary_key}"
+      Array.wrap(klass.primary_key).map { |pk| "#{klass.quoted_table_name}.#{pk}" }.join(', ')
```
magnusvk, line 179:
```suggestion
      Array.wrap(klass.quoted_primary_key).map { |pk| "#{klass.quoted_table_name}.#{pk}" }.join(', ')
```
I do think this should be `quoted_primary_key`. [According to ChatGPT](https://chatgpt.com/share/68422a52-8224-8006-b298-2934c520b0cf) that also returns an array, and should be fine to use here to preserve the previous quoting behavior?

lib/counter_culture/reconciler.rb
```diff
@@ -61,6 +61,7 @@ class Reconciliation
 
       delegate :model, :relation, :full_primary_key, :relation_reflect, :polymorphic?, :to => :counter
       delegate *CounterCulture::Counter::CONFIG_OPTIONS, :to => :counter
+      delegate :primary_key, :to => :relation_class
```
magnusvk, line 64: I got pretty confused by this `delegate`, particularly because there are also some local variables called `primary_key` that conflict with this. Can we remove that and just stick with the `relation_class.primary_key` calls? It's not that much more typing. 😄

lib/counter_culture/reconciler.rb
```diff
@@ -217,10 +220,23 @@ def count_select
             @count_select = "SUM(COALESCE(#{self_table_name}.#{delta_column}, 0))"
           end
         else
-          @count_select = "COUNT(#{self_table_name}.#{model.primary_key})*#{delta_magnitude}"
+          primary_key = Array.wrap(model.primary_key).first
```
magnusvk, line 223: Judging by the fact that the specs are green this seems to work, but my brain doesn't fully grok why it's ok to just use `.first` here. Any chance you can explain that to me? 😄

magnusvk (approved):
I think this looks good. It was a bit of a beast to review as it changes all of the core places in the gem, but... I think I got it, and the specs are green, so that's a good sign. If you don't mind just taking out the `delegate` I'll get this merged and released.

Thanks again for your work on this!

lib/counter_culture/reconciler.rb
notchairmk: Happy to remove. I think at the time I didn't think there was any shadowing? Idk, but was obviously wrong there. Should be good now!

lib/counter_culture/reconciler.rb
notchairmk: My thinking was that we just need a count of any column from the left join with the association. We really only need one column from the join, multiple counts will _should_ just return the same number. Here's an example of what the query looks like currently for one of the tests I added (including a duplicate select for `id`, but that's really minor and I think is an effect of selecting [both of these](https://github.com/magnusvk/counter_culture/blob/049113a1c876a295c74f4c06a52f7cbfbbee6d86/lib/counter_culture/reconciler.rb#L114-L115) in the master branch atm).

```sql
SELECT
    "composite_groups".id,
    "composite_groups".id,
    "composite_groups".secondary_id,
    COUNT("composite_group_users".id) * 1 AS count,
    MAX("composite_groups".composite_users_pt_count) AS composite_users_pt_count
FROM
    "composite_groups"
LEFT JOIN "composite_group_users" AS composite_group_users
    ON "composite_groups".id = composite_group_users.composite_group_id
        AND "composite_groups".secondary_id = composite_group_users.secondary_id
GROUP BY
    "composite_groups"."id"
ORDER BY
    "composite_groups"."id" ASC
LIMIT 1000
```

I'll admit that doing `Array.wrap(...).first` here in `count_select` but doing `Array.wrap(...).map` in `primary_key_select` feels inconsistent. I think the latter could be rewritten to match the former if you feel like the inconsistency is an issue. That'd just end up excluding `"composite_groups".secondary_id` from the above query.

Totally willing to adjust if something seems off with my line of thinking 😅

magnusvk: This does make sense to me, yeah, for this case.

I'm curious—does this addition also work if it ends up with a join that's the other way around? My instinct is that it might not?

I'm trying to compute this, but here, on this one-to-many, the composite key is on the many side, right? What if the composite key was on the one side? So if `composite_groups` had a composite primary key. Then wouldn't we have to group by all of those, and also count by all of them? Or would that still work?

magnusvk:
Ok, so one bit of delay here is that I'm going to be OOO and away from my computer for the next two weeks. I don't think it's a great idea to release this today and then not look at my Github notifications for two weeks in case something breaks, so I'm going to hold off on the merge until I'm back.

lib/counter_culture/reconciler.rb
magnusvk: Yeah, exactly.

notchairmk: I'd be happy to take a look, but I'm wondering if that syntax is supported outside of these composite primary key changes? I'm looking at the tests on the default branch and as far as I can tell all `counter_culture`s are defined on the child.

Making this change locally, tests start failing on the default branch (unrelated to the reconciler):
```diff
diff --git a/spec/models/review.rb b/spec/models/review.rb
index 269df2a..6e4ab49 100644
--- a/spec/models/review.rb
+++ b/spec/models/review.rb
@@ -4,7 +4,6 @@ class Review < ActiveRecord::Base
 
   counter_culture :product, :touch => true, :with_papertrail => PapertrailSupport.supported_here?
   counter_culture :product, :column_name => 'rexiews_count', touch: :rexiews_updated_at
-  counter_culture :user
   counter_culture :user, :column_name => proc { |model| model.review_type && model.review_type != 'null' ? "#{model.review_type}_count" : nil }, :column_names => {"reviews.review_type = 'using'" => 'using_count', "reviews.review_type = 'tried'" => 'tried_count', "reviews.review_type = 'null'" => nil}
   counter_culture :user, :column_name => 'review_approvals_count', :delta_column => 'approvals'
   counter_culture :user, :column_name => 'review_value_sum', :delta_column => 'value'
diff --git a/spec/models/user.rb b/spec/models/user.rb
index b040cc6..3ee861e 100644
--- a/spec/models/user.rb
+++ b/spec/models/user.rb
@@ -12,6 +12,7 @@ class User < ActiveRecord::Base
 
   has_many :reviews
   accepts_nested_attributes_for :reviews, :allow_destroy => true
+  counter_culture :reviews, :column_name => 'reviews_count'
 
   if PapertrailSupport.supported_here?
     has_paper_trail

```

Interested to hear your thoughts on how to proceed. Like I said I can take a look if that's your preference, but it feels like it'd potentially be handling it differently than how it's handled in tests currently.

magnusvk: Yeah, this actually occurred to me while I was walking my dog yesterday. I think you're right, this just isn't something the gem envisions, and also really doesn't make a ton of sense, because if you're counting a one-to-many association, you want to count the many side.

Ok, thank you for entertaining me here, I just had a hard time loading up all that complexity. I think we're good on this.

magnusvk:
Just released as gem version 3.11.0. Thanks for the contribution!

jfloff:
I’m sorry to dig up this thread, but after updating counter_culture to this version, my setup no longer works as expected and I could use some help figuring out why. Here’s what’s happening:

I have this on my `Item` model:

```ruby
counter_culture :category, column_name: 'items_count', foreign_key_values: proc { |category_id|
  [
    category_id,
    ItemCategory.find_by(id: category_id).try(:parent).try(:id)
  ].compact
}
```

... and on ItemCategory **using the [ancestry gem](https://github.com/stefankroes/ancestry)**:

```ruby
has_many   :items, dependent: :restrict_with_error, foreign_key: :category_id
has_ancestry

```

I expect each category’s `items_count` to include items assigned directly to it and items assigned to its immediate children -- so adding an item to a subcategory should bump both the child’s and the parent’s counts. 
Instead, only the child’s count is changing. It feels like the second ID returned by my `foreign_key_values` proc is being ignored.

I’ve confirmed that `parent_id` is set correctly and that `ItemCategory.find(...).parent.id` returns the right value. Has something changed that might affect this specific use case?

Thank you,

magnusvk:
Hmm. @notchairmk any ideas here? Also, @jfloff any chance you can put up a PR with a minimal failing test case that reproduces this?

jfloff:
Sorry its kinda late here 😅  so I won't have time for the PR, but I will just dump my current findings:

The problem seems to at least materialize here:
```
conditions = primary_key_conditions(primary_key, id_to_change)
```

In previous versions the `id_to_change` would show up like this `[3, 1]` but in the new version it shows up like this `[[3], 1]` which then on the `zip` within `primary_key_conditions` gets lost, and we’re left only with the first key.

Now the problem is that even if we have `[3, 1]`, with the current zip approach, since we only have a single `primary_key` we end up losing the second value: `["id"].zip([3,1])  # => [["id", 3]]`.

I think the goal was for the `id_to_change` input to be `[[3,1]]` that way when you zip `["id"].zip([[3,1]])` you get `[["id", [3, 1]]]`.

But when we call the `foreign_key_values` proc, with an `id_to_change` like `[3]`, my proc would return `[[3], 1]`. 

I think the goal here is more or less to:
1) flatten the output of `foreign_key_values` proc
2) wrap it in another array

Either this or we the `primary_key_conditions` has to change to accommodate outputs like `[[3], 1]`

magnusvk:
Could you fix this by changing what your proc returns to be the flattened array?

notchairmk:
Sorry, just catching up on this. Is the workaround you have fully working @jfloff?

@magnusvk depending on how you're feeling about how the library should have opinions based on a mismatch between primary key and foreign key cardinality (1 primary key, N foreign keys), I can take a look. Specifically in this case it seems like `primary_key_conditions` is where we'd have to change:
https://github.com/magnusvk/counter_culture/blob/612d4f5aaa919b27b34195757a3f6766f7de9dcd/lib/counter_culture/counter.rb#L449-L453
We have a singular `primary_key` and an array of `fk_value`, so could check if that's the case and could effectively say "if you have on PK and an array of FK, then we assume that means that you want to map all FKs to the one PK value." I'd personally be somewhat hesitant, but can take a look at it if you think that'd be something to support in the library.

magnusvk:
See #417 for more—I do think this is a regression we should fix, sorry for breaking things. I would love a review from the two of you on this though before I release it.

Issue #414, opened 2025-06-12
Counter cache is not updated in memory when belongs_to association is changed without reload


yuma-ito-bd:
## Background

There are two models, `SimpleMain` and `SimpleDependent`, with a `has_many` / `belongs_to` relationship:

```ruby
class SimpleMain < ActiveRecord::Base
  has_many :simple_dependents
end

class SimpleDependent < ActiveRecord::Base
  belongs_to :simple_main

  counter_culture :simple_main
end
```

## Issue

When the `belongs_to` association of a `SimpleDependent` instance is changed, the counter cache on the new parent (`SimpleMain`) is not updated unless `reload` is called. This means the in-memory value is stale until reloaded from the database.

## How to Reproduce

Add the following test to `counter_culture_spec.rb`, and it fails:

```ruby
it "should update belongs_to counter caches without reload" do
  simple_main1 = SimpleMain.create
  simple_main2 = SimpleMain.create
  simple_dependent = SimpleDependent.create :simple_main_id => simple_main1.id

  simple_dependent.simple_main = simple_main2
  simple_dependent.save!

  expect(simple_dependent.simple_main.simple_dependents_count).to eq(1)
end
```

### Test Output

```
Failures:

  1) CounterCulture should not fix belongs_to counter caches
     Failure/Error: expect(simple_main2.simple_dependents_count).to eq(1)
     
       expected: 1
            got: 0
     
       (compared using ==)
     # ./spec/counter_culture_spec.rb:1385:in 'block (2 levels) in <top (required)>'
```

This indicates that after changing the association, the counter cache is not updated in memory unless `reload` is called.

Thank you.

PR #415, opened 2025-06-14
fix: Skip assigning counter to associated objects when referencing past values

yuma-ito-bd:
fixed: https://github.com/magnusvk/counter_culture/issues/414

When `options[:was]` is true (referencing past values), skip assigning the counter value to associated objects. 
This prevents unnecessary counter assignments when referencing past values.


vitalinfo (approved):

magnusvk:
Thank you, released as gem version 3.10.2.

yuma-ito-bd:
@magnusvk 
Thank you so much!

PR #416, opened 2025-06-27
fix: FrozenError when associated instance is frozen

r-plus:
Hi @magnusvk 

After the v3.10.0 of this change https://github.com/magnusvk/counter_culture/pull/407 we encountered the `FrozenError:
       can't modify frozen attributes` error while the create/destroy. Because we have frozen associated object instance while counter_culture attempt update the attribute of instance.

This pull-request fix the issue and add test about it.


magnusvk:
Thank you! Released as gem version 3.11.1.

jeffbax:
Thank you for fixing this!

PR #417, opened 2025-07-08
Fix regression for `foreign_key_values` by adding support for composite primary keys

magnusvk:
It turns out the old tests here weren't properly reloading and thus didn't catch this regression.


notchairmk (approved):
Thanks for fixing this!

jfloff:
@magnusvk I was on vacation, sorry for just catching up to this :) Looks great to me! Thank you for being on top of this.

PR #418, opened 2025-08-20
Demonstrate broken counter in after update callback

badlamer:
Hello! 
I found a problem with updating the counter when the record is saved in the after_update callback.

The counter is updated twice because _update_counts_after_update is called twice

I added a sample test in the PR


magnusvk:
So my first instinct is to say "it seems pretty weird to call `save` in the `after_update` for that same model—no wonder this breaks!" If that was my code I would just move that update to a `before_validation` hook and save the extra database update.

That being said, if you could figure out an approach to fix this I would be happy to look at a PR for this! I won't have time to proactively fix this myself unfortunately.

PR #420, opened 2025-10-22
Correctly update in-memory counter when using conditional counter caches

### LINKED ISSUE #419: Conditional decrementing not updating record in memory, only when incrementing
The following spec fails:

```rb
  it 'should correctly increment and decrement counters of associated objects' do
    conditional_main = ConditionalMain.create
    conditional_dependent = conditional_main.conditional_dependents.create(condition: false)

    expect(conditional_main.conditional_dependents_count).to eq(0)

    conditional_dependent.update(condition: true)

    # works correctly, value is correct without reload
    expect(conditional_main.conditional_dependents_count).to eq(1)

    conditional_dependent.update(condition: false)

    # doesn't work, value is incorrect without reload
    expect(conditional_main.conditional_dependents_count).to eq(0)
  end
```

So incrementing the value works, but not decrementing it.

I traced it back to #415. Before this change it works as expected, so the bug was introduced by it.

magnusvk:
Fixes #419

### DIFF
--- a/spec/counter_culture_spec.rb
+++ b/spec/counter_culture_spec.rb
@@ -140,7 +140,6 @@ RSpec.describe "CounterCulture" do
 
     conditional_dependent.update(condition: true)
 
-    # works correctly, value is correct without reload
     expect(conditional_main.conditional_dependents_count).to eq(1)
 
     conditional_dependent.update(condition: false)


spec/counter_culture_spec.rb
```diff
@@ -132,6 +132,23 @@ def yaml_load(yaml)
...
 
+  it "should correctly increment and decrement counters of associated objects" do
+    conditional_main = ConditionalMain.create
+    conditional_dependent = conditional_main.conditional_dependents.create(condition: false)
+
+    expect(conditional_main.conditional_dependents_count).to eq(0)
+
+    conditional_dependent.update(condition: true)
+
+    # works correctly, value is correct without reload
```
doits:
```suggestion
```

spec/counter_culture_spec.rb
```diff
@@ -132,6 +132,23 @@ def yaml_load(yaml)
...
+    expect(conditional_main.conditional_dependents_count).to eq(1)
+
+    conditional_dependent.update(condition: false)
+
+    # doesn't work, value is incorrect without reload
```
doits:
```suggestion
```

lib/counter_culture/counter.rb
```diff
@@ -150,7 +150,27 @@ def change_counter_cache(obj, options)
...
+            # keys are identical, we're in scenario 2 and should update the in-memory counter.
+            should_update_counter = if options[:was]
+              current_fk = foreign_key_value(obj, relation, false)
+              previous_fk = foreign_key_value(obj, relation, true)
+              current_fk == previous_fk && current_fk.present?
+            else
+              true
+            end
+
+            if should_update_counter
```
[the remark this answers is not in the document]
magnusvk: Yeah, I see what you're saying, but the problem is that the method would have to take more than just `options`—it needs `obj` and `relation` as well. And it's only used in the one place. So I'll leave this as-is for now.

Thanks for the suggestion!

PR #421, opened 2025-11-01
Fix counter cache not using the correct type with multiple STI models in the association chain

andriusch:
Before this PR `CounterCulture::Reconciler` didn't handle having STI models in the middle of association chain correctly if the model that defines counter cache is also STI model. When generating join queries it would use `type = "[model.type]"`, but `[model.type]` would always be the type of the model that called `counter_culture_fix_counts`.

For example:
```ruby
class AdminUser < User
  belongs_to :manages_company
end

class TwitterReview < Review
  belongs_to :admin_user
  counter_culture [:admin_user, :manages_company]
```

The first join will be generated incorrectly:
```SQL
FROM companies
LEFT JOIN users ON companies.id = users.managed_company_id AND users.type = "TwitterReview"
LEFT JOIN reviews ON users.id = reviews.admin_user_id AND reviews.type = "TwitterReview"
```

This PR fixes it by setting correct `model` for each relation, so the same query becomes:

```SQL
FROM companies
LEFT JOIN users ON companies.id = users.managed_company_id AND users.type = "AdminUser"
LEFT JOIN reviews ON users.id = reviews.admin_user_id AND reviews.type = "TwitterReview"
```

Since it sets `model` local variable in `Reconciler`, this should also fix intermediate models that were `paranoia` or `discard` deleted, but would still be joined in the join query, though I don't use either so I haven't tested this interaction.

### DIFF
--- a/lib/counter_culture/reconciler.rb
+++ b/lib/counter_culture/reconciler.rb
@@ -59,7 +59,7 @@ module CounterCulture
     class Reconciliation
       attr_reader :counter, :options, :relation_class
 
-      delegate :model, :relation, :full_primary_key, :relation_reflect, :polymorphic?, :to => :counter
+      delegate :model, :relation, :full_primary_key, :relation_reflect_and_model, :relation_reflect, :polymorphic?, :to => :counter
       delegate *CounterCulture::Counter::CONFIG_OPTIONS, :to => :counter
 
       def initialize(counter, changes_holder, options, relation_class)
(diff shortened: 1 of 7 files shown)


lib/counter_culture/reconciler.rb
```diff
@@ -299,8 +299,8 @@ def join_clauses(where)
 
           # adds 'type' condition to JOIN clause if the current model is a
           # child in a Single Table Inheritance
-          if reflect.active_record.column_names.include?('type') &&
-              !model.descends_from_active_record?
+          if model.column_names.include?('type') &&
```
andriusch: I think it should be using `model.inheritance_column` instead of `'type'`, but I've never seen anyone using any other inheritance column aside from `type`, so probably not a real problem, though I could also fix it in this PR.

lib/counter_culture/reconciler.rb
```diff
@@ -257,11 +257,11 @@ def join_clauses(where)
...
         # conditions
         join_clauses = reverse_relation.each_with_index.map do |cur_relation, index|
-          reflect = relation_reflect(cur_relation)
+          reflect, model = relation_reflect_and_model(cur_relation)
 
-          target_table = quote_table_name(reflect.active_record.table_name)
+          target_table = quote_table_name(model.table_name)
           target_table_alias = parameterize(target_table)
-          if relation_class.table_name == reflect.active_record.table_name
+          if relation_class.table_name == model.table_name
```
magnusvk: Am I reading this right that these `reflect.active_record` -> `model` changes don't change behavior, they just take the short-cut straight to `model` now that we have it here?

magnusvk: Makes sense, thank you.

PR #429, opened 2026-04-17
Fix StaleObjectError when counter target model has optimistic locking

savin-alexey:
## Problem

Since v3.12.1 (#425), counter_culture uses hash-based `update_all` for Arel compatibility. Rails auto-increments `lock_version` when `update_all` receives a Hash and the model has optimistic locking - see   [ActiveRecord::Relation#update_all](https://github.com/rails/rails/blob/v8.1.3/activerecord/lib/active_record/relation.rb#L615-L621).

Before v3.12.1 counter_culture used string-based `update_all` which Rails does not modify.

So now counter updates silently bump `lock_version` in the DB while the in-memory object keeps the old value. If the parent also has `belongs_to ... touch: true` children, the next `touch` finds a stale `lock_version` and raises `ActiveRecord::StaleObjectError`.

## Fix

Include `lock_version = lock_version` (no-op) in the `arel_updates` hash before calling `update_all`. Rails sees the key is already there and skips the auto-increment.

Counter updates are atomic SQL (`SET col = col + 1`) - they do not need optimistic locking.

Applied to both immediate and aggregate update paths.

### DIFF
--- a/lib/counter_culture.rb
+++ b/lib/counter_culture.rb
@@ -46,8 +46,7 @@ module CounterCulture
         end
 
         if arel_updates.any?
-          # Prevent Rails 8.1+ from auto-incrementing lock_version
-          # (same rationale as in Counter#change_counter_cache)
+          # Set lock_version = lock_version (no-op) to skip Rails auto-increment
           if klass.locking_enabled?
             lc = klass.locking_column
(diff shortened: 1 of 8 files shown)


magnusvk:
Thanks for the fix! Just released as gem version 3.13.1.

Issue #431, opened 2026-06-25
Multi-level counter cache is not updated in memory


baarde:
Hello,

## Issue

When creating or deleting a new record, single-level counter caches are updated in memory but multi-level counter caches are not.

## How to Reproduce

Add the following test:

```ruby
it "increments multi-level counter cache on create without reload" do
  company = Company.create
  user = User.create manages_company: company

  expect { user.reviews.create }.to change(company, :reviews_count).by(1)
end
```

### Test Output

```
Failures:

  1) CounterCulture basic counter cache increments multi-level counter cache on create without reload
     Failure/Error: expect { user.reviews.create }.to change(company, :reviews_count).by(1)
       expected `Company#reviews_count` to have changed by 1, but was changed by 0
     # ./spec/counter_culture/basic_counter_cache_spec.rb:77:in 'block (2 levels) in <top (required)>'
```

PR #432, opened 2026-06-25
Update multi-level counter cache in memory

baarde:
Hello,

This PR is a fix for #431.

Currently, updating a multi-level counter cache in memory doesn't work because `#association_object_for_assign` cannot access the leaf association from the root object.

I've modified the method to traverse the chain of associations and return the leaf object that holds the in-memory counter cache (unless we encounter `nil` or an un-loaded association along the way).


magnusvk:
Thank you! Released as gem version 3.13.2.

Issue #263, opened 2019-07-01
Support Updating Counter Cache After Commit (to avoid deadlocks)


jbritten:
Support updating the counter cache after commit (outside the primary transaction) where the SQL UPDATE calls would be less susceptible to deadlocks.

Let's say we've got a Campaign model which can have many Subscribers. We store lots of campaign metrics in a separate CampaignMetrics summary table. When a subscriber confirms their email address we may see a MySQL transaction such as the following, which illustrates counter_culture updating various counts as well as additional proprietary work being performed:

```sql
BEGIN
SELECT 1 AS one FROM `subscribers` WHERE (`subscribers`.`email` = 'user@host.com' AND `subscribers`.`campaign_id` = 1001) LIMIT 1
SELECT `campaigns`.* FROM `campaigns` WHERE `campaigns`.`id` = 1001 LIMIT 1
UPDATE `subscribers` SET `confirmed_at` = '2019-07-01 16:34:04', `status` = 'confirmed', `updated_at` = '2019-07-01 16:34:04' WHERE `subscribers`.`id` = 2001
SELECT `campaign_metrics`.* FROM `campaign_metrics` WHERE `campaign_metrics`.`campaign_id` = 1001 LIMIT 1
UPDATE `campaign_metrics` SET `confirmed_subscribers_count` = COALESCE(`confirmed_subscribers_count`, 0) + 1 WHERE `campaign_metrics`.`id` = 1001
SELECT `campaigns`.* FROM `campaigns` WHERE `campaigns`.`id` = 1001 LIMIT 1
SELECT `campaign_metrics`.* FROM `campaign_metrics` WHERE `campaign_metrics`.`campaign_id` = 1001 LIMIT 1
UPDATE `campaign_metrics` SET `unconfirmed_subscribers_count` = COALESCE(`unconfirmed_subscribers_count`, 0) - 1 WHERE `campaign_metrics`.`id` = 1001
 ... (some additional proprietary work) ...
COMMIT
```

There are other actions a Subscriber could do which will also update data in the CampaignMetrics table. For example, opening an email received would update the email_opens_count on the CampaignMetrics table with a transaction such as the following:

```sql
BEGIN
SELECT 1 AS one FROM `subscribers` WHERE (`subscribers`.`email` = 'user@host.com' AND `subscribers`.`campaign_id` = 1001) LIMIT 1
[...]

PR #435, opened 2026-06-26
Use ActiveRecord.after_all_transactions_commit on Rails 7.2+

rainerborene:
## What

`execute_after_commit` currently always depends on the [`after_commit_action`](https://github.com/comboy/after_commit_action) gem to defer counter cache updates until after the surrounding transaction commits. Rails 7.2 introduced [`ActiveRecord.after_all_transactions_commit`](https://github.com/rails/rails/pull/51474), which does exactly the same thing natively.

This PR prefers the built-in API when it's available, so on Rails 7.2+ `execute_after_commit` no longer requires any extra dependency. On older Rails versions nothing changes — it keeps using `after_commit_action` exactly as before.

## Why

This is the recurring request in #263 and #120 (avoiding deadlocks by moving counter updates after commit). The original reason for pulling in a separate gem was that Rails had no public hook for "run this after the outermost transaction commits." Rails 7.2 now ships one, so on modern Rails the dependency is redundant.

As a small bonus, `after_all_transactions_commit` runs the block **immediately** when there is no open transaction, whereas the per-record `after_commit` hook would otherwise never fire in that case.

## Changes

- `Counter#initialize`: only `require 'after_commit_action'` / `include AfterCommitAction` when `ActiveRecord` does **not** respond to `after_all_transactions_commit` (i.e. Rails < 7.2).
- `Counter#execute_now_or_after_commit`: dispatch through `ActiveRecord.after_all_transactions_commit` when available, falling back to `obj.execute_after_commit`.
- Specs: added a context (guarded by `ActiveRecord.respond_to?(:after_all_transactions_commit)`) asserting the update is routed through the native API and that the `after_commit_action` module is not mixed into the model on modern Rails.
- README + CHANGELOG updated.

## Testing

Full suite passes on Rails 8.1 / sqlite3 (`218 examples, 0 failures`, 2 PostgreSQL-only pending). The existing `execute_after_commit` behavior specs already cover the deferral semantics and now exercise the native path on Rails 7.2+; the gem fallback path remains covered by the older-Rails entries in the CI matrix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+New features:
+  - Use the built-in `ActiveRecord.after_all_transactions_commit` API for `execute_after_commit` on Rails 7.2+, so the `after_commit_action` gem is no longer required on modern Rails. Older Rails versions keep using the gem as before.
+
 ## 3.13.3 (June 25, 2026)
 
 Bugfixes:
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ Another option is to simply defer the update of counter caches to outside of the
 ```ruby
   counter_culture :category, execute_after_commit: true
 ```
-[NOTE] You need to manually specify the `after_commit_action` as dependency in the Gemfile to use this feature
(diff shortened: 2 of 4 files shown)


magnusvk:
Thank you! I merged this in #436 with just a couple of small tweaks to avoid the back and forth. Released as gem version 3.14.0 just now.

